PR #7, opened 2016-10-11
Improvement on the Language Chooser

### LINKED ISSUE #4: Don't add "use" to language
In the spec, there's a comment that I think is correct. Adding "Use" doesn't really add any value. It just makes the list harder to scan and adds something else to translate.

tintou:
* New animation for the Language Chooser
* Now only using the language name and translate it into the language. Fixes #4


tintou:
I've fixed it by giving it a 64px height

PR #12, opened 2016-10-28
Add DiskView and CheckView

tintou:
DiskView:
 - Get all the existing OS
 - Get it for different disks

CheckView:
 - Check for Disk Space
 - Check for Processor & Memory
 - Check for Battery


danrabbit:
- [ ] There's a comment in the doc that instead of "Skip" we should use "Ignore" since this is a bit more clear that we're ignoring this message

- [ ] "Erase and Install" should be a destructive action class since it involves the loss of data

PR #19, opened 2016-10-29
fix the language heading margin

btkostner:
This fixes some margin things on the language selector


tintou:
16 is not a multiple of 3…

PR #41, opened 2017-07-02
Look for .VolumeIcon.png in partition roots.

vjr:
This is to address issue https://github.com/elementary/pantheon-installer/issues/39

See https://github.com/dab2000/rEFInd-custom/blob/master/refind/lib.c#L958 and https://github.com/dab2000/rEFInd-custom/blob/master/libeg/image.c#L366 for similar code in rEFInd.

Note that in the `Pixbuf.from_file_at_scale ()` call I've hardcoded the height value to 96 which makes it match the other icons in my installer screen. I need advice on whether that is OK or is there a good way to match the ThemedIcon size being set?

Screenshot where it shows this picking up my KDE Neon partition icon instead of displaying the default linux tux icon:

installer


vjr:
Note the installer falls back to showing the included default linux tux icon if the volume icon png is not found, like you can see in this screenshot for the KDE Neon entry:

installer_old

vjr:
Should we look for only `.VolumeIcon.png` or just `.VolumeIcon` or both filenames?

danrabbit:
I think in this case you want to instead use [GLib.FileIcon](https://valadoc.org/gio-2.0/GLib.FileIcon.html)

Issue #53, opened 2017-08-28
Use system-os-installer instead of system-os-install


cassidyjames:
Depends on https://github.com/elementary/icons/issues/336

danrabbit:
Fixed by #54

PR #54, opened 2017-08-28
Change 'system-os-install' to 'system-os-installer' to be generic

isantop:
The `system-os-install` icon doesn't seem to be standardized, whereas `system-os-installer` is used by a large number of OS installers.


isantop:
Fixes #53

PR #63, opened 2017-08-28
Subclass Gtk.Window instead of Dialog

### LINKED ISSUE #62: Move to window instead of dialog
Not supposed to have a parent-less dialog. Instead we should make it a window with flat, rounded, and default-decoration style classes.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/48738796-move-to-window-instead-of-dialog?utm_campaign=plugin&utm_content=tracker%2F46189108&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F46189108&utm_medium=issues&utm_source=github).
</bountysource-plugin>

danrabbit:
Gets rid of a warning about no transient parent

Fixes #62

### DIFF
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -18,7 +18,7 @@
  * Authored by: Corentin Noël <corentin@elementary.io>
  */
 
-public class Installer.MainWindow : Gtk.Dialog {
+public class Installer.MainWindow : Gtk.Window {
     private CheckView check_view;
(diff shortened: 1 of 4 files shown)


cassidyjames:
Never mind

PR #72, opened 2017-08-29
Set dialog title to OS release

### LINKED ISSUE #64: Set title to "Installing OS_NAME"
...with `OS_NAME` being the name of the OS from LSB.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/48739283-set-title-to-installing-os_name?utm_campaign=plugin&utm_content=tracker%2F46189108&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F46189108&utm_medium=issues&utm_source=github).
</bountysource-plugin>

danrabbit:
fixes #64

### DIFF
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -76,18 +76,12 @@ public class Installer.MainWindow : Gtk.Dialog {
         const string ETC_OS_RELEASE = "/etc/os-release";
 
         try {
-            var osrel = new Gee.HashMap<string, string> ();
             var data_stream = new DataInputStream (File.new_for_path (ETC_OS_RELEASE).read ());
 
             string line;
             while ((line = data_stream.read_line (null)) != null) {
                 var osrel_component = line.split ("=", 2);
-
-                if (osrel_component.length == 2) {
-                    osrel[osrel_component[0]] = osrel_component[1].replace ("\"", "");
-                }
-
-                if (osrel_component[0] == "PRETTY_NAME") {
+                if (osrel_component.length == 2 && osrel_component[0] == "PRETTY_NAME") {
                     pretty_name = osrel_component[1].replace ("\"", "");
                     break;
                 }
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -38,6 +38,7 @@ public class Installer.MainWindow : Gtk.Dialog {
         stack.add_named (keyboard_layout_view, "keyboard-layout");
         stack.add_named (progress_view, "progress-view");
 
+        title = _("Install %s").printf (get_os_release ());
         set_default_geometry (800, 600);
         get_content_area ().add (stack);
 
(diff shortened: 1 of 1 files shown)


src/MainWindow.vala
```diff
@@ -69,4 +70,29 @@ public class Installer.MainWindow : Gtk.Dialog {
             stack.set_visible_child_name ("progress-view");
         });
     }
+
+    private string get_os_release () {
```
donadigo: This method should be static as it doesn't access anything from the class instance.

src/MainWindow.vala
```diff
@@ -69,4 +70,29 @@ public class Installer.MainWindow : Gtk.Dialog {
...
+        string os;
+        var file = File.new_for_path("/etc/os-release");
```
donadigo: Can this `/etc/os-release` be a constant?

Also, spaces between braces `new_for_path ()`.

src/MainWindow.vala
```diff
@@ -69,4 +70,29 @@ public class Installer.MainWindow : Gtk.Dialog {
...
+
+        try {
+            var osrel = new Gee.HashMap<string, string> ();
```
donadigo: If this is the only information we're gonna get from that file, this `HashMap` isn't needed here and we can just assign the OS string in the `while` statement and then `break` after we're done. If that's not the case we should probably make a new class to parse and provide the needed information about the OS to not clutter the `MainWindow` class.

src/MainWindow.vala
Conan-Kudo: This is basically constant. This path never changes.

src/MainWindow.vala
```diff
@@ -69,4 +70,29 @@ public class Installer.MainWindow : Gtk.Dialog {
...
+                var osrel_component = line.split ("=", 2);
+                if ( osrel_component.length == 2 ) {
+                    osrel[osrel_component[0]] = osrel_component[1].replace ("\"", "");
+                }
+            }
+
+            os = osrel["PRETTY_NAME"];
+        } catch (Error e) {
+            warning("Couldn't read os-release file, assuming elementary OS");
+            os = "elementary OS";
```
Conan-Kudo: I would suggest assuming `Generic OS` or something unbranded.

src/MainWindow.vala
Conan-Kudo: This is basically constant. This path never changes.

donadigo: I mean, to have this declared in an a `const string`.

Conan-Kudo: It should be a `const string` because it's basically constant, I mean. :)

PR #77, opened 2017-08-29
Try or Install View

isantop:
Adds a new pane to allow the user to either try or install the OS. Try will drop them to a live session, whereas install will proceed with the installation.

### DIFF
--- a/src/Views/TryInstallView.vala
+++ b/src/Views/TryInstallView.vala
@@ -19,12 +19,14 @@
 public class TryInstallView : AbstractInstallerView {
     public signal void next_step ();
 
+    // replace this with actual content when the type is ready.
+
     construct {
         var title_label = new Gtk.Label (_("Welcome"));
         title_label.get_style_context ().add_class ("h2");
         title_label.valign = Gtk.Align.START;
 
-        var description_label = new Gtk.Label (_("You can try out OS_NAME before you install it.\n\nReady to take the plunge? Click the Install button."));
+        var description_label = new Gtk.Label (_("You can try out OS_NAME before you install it. If you're ready, you can install it any time."));
         title_label.valign = Gtk.Align.START;
 
         var image = new Gtk.Image.from_icon_name ("distributor-logo", Gtk.IconSize.DIALOG);
@@ -32,7 +34,6 @@ public class TryInstallView : AbstractInstallerView {
 
         var selection_grid = new Gtk.Grid ();
         selection_grid.orientation = Gtk.Orientation.VERTICAL;
-        selection_grid.add (title_label);
         selection_grid.add (description_label);
         selection_grid.add (image);
 
@@ -41,8 +42,9 @@ public class TryInstallView : AbstractInstallerView {
         content_area.margin_start = 10;
         content_area.attach (image, 0, 0, 1, 1);
         content_area.attach (title_label, 0, 1, 1, 1);
+        content_area.attach (selection_grid, 0, 1, 1, 1);
 
-        var cancel_button = new Gtk.Button.with_label (_("Try out OS_NAME"));
+        var cancel_button = new Gtk.Button.with_label (_("Cancel installation"));
 
         var next_button = new Gtk.Button.with_label (_("Install OS_NAME"));
         next_button.get_style_context ().add_class (Gtk.STYLE_CLASS_SUGGESTED_ACTION);
--- a/src/Views/TryInstallView.vala
+++ b/src/Views/TryInstallView.vala
@@ -34,14 +34,23 @@ public class TryInstallView : AbstractInstallerView {
         nochanges_label.valign = Gtk.Align.START;
         nochanges_label.wrap = true;
 
-        var nochanges_image = new Gtk.Image.from_icon_name ("checkmark", Gtk.IconSize.DIALOG);
+        var nochanges_image = new Gtk.Image.from_icon_name ("computer", Gtk.IconSize.DIALOG);
         nochanges_image.valign = Gtk.Align.START;
+        var nochanges_emblem = new Gtk.Image.from_icon_name ("emblem-default", Gtk.IconSize.LARGE_TOOLBAR);
+        nochanges_emblem.valign = Gtk.Align.END;
+        nochanges_emblem.halign = Gtk.Align.END;
+
+        var nochanges_overlay = new Gtk.Overlay ();
+        nochanges_overlay.add_overlay (nochanges_image);
+        nochanges_overlay.add_overlay (nochanges_emblem);
+        nochanges_overlay.halign = Gtk.Align.CENTER;
+        nochanges_overlay.width_request = 48;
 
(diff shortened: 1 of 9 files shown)


danrabbit:
Depends on #72

src/Views/TryInstallView.vala
```diff
@@ -0,0 +1,99 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+public class TryInstallView : AbstractInstallerView {
+    public signal void next_step ();
+
+    // replace this with actual content when the type is ready.
```
danrabbit: Remove this comment please

src/Views/TryInstallView.vala
```diff
@@ -0,0 +1,99 @@
...
+        description_label.valign = Gtk.Align.START;
+        description_label.hexpand = true;
+
+        var nochanges_label = new Gtk.Label (_("Data from your previous operating system is unchanged until you install OS"));
+        nochanges_label.valign = Gtk.Align.START;
+        nochanges_label.wrap = true;
+
+        var nochanges_image = new Gtk.Image.from_icon_name ("computer", Gtk.IconSize.DIALOG);
+        nochanges_image.valign = Gtk.Align.START;
+        var nochanges_emblem = new Gtk.Image.from_icon_name ("emblem-default", Gtk.IconSize.LARGE_TOOLBAR);
```
danrabbit: newline before this line

src/Views/TryInstallView.vala
```diff
@@ -0,0 +1,99 @@
...
+        nochanges_emblem.valign = Gtk.Align.END;
+        nochanges_emblem.halign = Gtk.Align.END;
+
+        var nochanges_overlay = new Gtk.Overlay ();
+        nochanges_overlay.add_overlay (nochanges_image);
+        nochanges_overlay.add_overlay (nochanges_emblem);
+        nochanges_overlay.halign = Gtk.Align.CENTER;
+        nochanges_overlay.width_request = 48;
```
danrabbit: Can we set this to 60 instead? Otherwise the icon isn't fully moved into the corner since the canvas has "matting"

src/Views/TryInstallView.vala
```diff
@@ -0,0 +1,99 @@
...
+        var nochanges_image = new Gtk.Image.from_icon_name ("computer", Gtk.IconSize.DIALOG);
+        nochanges_image.valign = Gtk.Align.START;
+        var nochanges_emblem = new Gtk.Image.from_icon_name ("emblem-default", Gtk.IconSize.LARGE_TOOLBAR);
+        nochanges_emblem.valign = Gtk.Align.END;
+        nochanges_emblem.halign = Gtk.Align.END;
+
+        var nochanges_overlay = new Gtk.Overlay ();
+        nochanges_overlay.add_overlay (nochanges_image);
+        nochanges_overlay.add_overlay (nochanges_emblem);
+        nochanges_overlay.halign = Gtk.Align.CENTER;
```
danrabbit: Can we add a height request of 51 here for the same reason as the next comment?

src/Views/TryInstallView.vala
```diff
@@ -0,0 +1,99 @@
...
+        nochanges_overlay.width_request = 48;
+
+        var nosaving_label = new Gtk.Label (_("Any changes you make in OS before installing will not be saved"));
```
danrabbit: This needs to use `printf` with the OS name as a variable. For example:

`_("Any changes you make in %s before installing will not be saved").printf (os_name);`

src/Views/TryInstallView.vala
```diff
@@ -0,0 +1,99 @@
...
+        nosaving_label.valign = Gtk.Align.START;
+        nosaving_label.wrap = true;
+
+        var nosaving_image = new Gtk.Image.from_icon_name ("document-revert", Gtk.IconSize.DIALOG);
+        nosaving_image.valign = Gtk.Align.START;
```
danrabbit: It seems like you're setting the valign for these icons opposites of each other. It doesn't seem necessary to set it at all

src/Views/TryInstallView.vala
```diff
@@ -0,0 +1,99 @@
...
+
+        var help_grid = new Gtk.Grid ();
+        help_grid.orientation = Gtk.Orientation.VERTICAL;
+        help_grid.column_homogeneous = true;
+        help_grid.vexpand = true;
+        help_grid.margin_end = 60;
+        help_grid.margin_start = 60;
+        help_grid.row_spacing = 10;
+        help_grid.column_spacing = 24;
+        help_grid.margin_top = 150;
```
danrabbit: No magic numbers. If the design calls for floating this grid center, then you should valign center

src/Views/TryInstallView.vala
```diff
@@ -0,0 +1,99 @@
...
+        return_label.wrap = true;
+
+        var return_image = new Gtk.Image.from_icon_name ("system-os-installer", Gtk.IconSize.DIALOG);
+        return_image.valign = Gtk.Align.END;
+
+        var help_grid = new Gtk.Grid ();
+        help_grid.orientation = Gtk.Orientation.VERTICAL;
+        help_grid.column_homogeneous = true;
+        help_grid.vexpand = true;
+        help_grid.margin_end = 60;
```
danrabbit: No magic numbers. If the design calls for floating this grid center, then you should halign center.

src/Views/TryInstallView.vala
```diff
@@ -0,0 +1,99 @@
...
+        help_grid.margin_start = 60;
+        help_grid.row_spacing = 10;
```
danrabbit: spacing should be a multiple of 3 (so 12)

src/Views/TryInstallView.vala
```diff
@@ -0,0 +1,99 @@
...
+        help_grid.margin_top = 150;
+
+        help_grid.attach (nochanges_overlay, 0, 0, 1, 1);
+        help_grid.attach (nochanges_label, 0, 1, 1, 1);
+        help_grid.attach (nosaving_image, 1, 0, 1, 1);
+        help_grid.attach (nosaving_label, 1, 1, 1, 1);
+        help_grid.attach (return_image, 2, 0, 1, 1);
+        help_grid.attach (return_label, 2, 1, 1, 1);
+
+        content_area.column_homogeneous = true;
```
danrabbit: Since you don't have any columns in this grid, this seems unnecessary

src/Views/TryInstallView.vala
```diff
@@ -0,0 +1,99 @@
...
+        content_area.margin_top = 10;
+        content_area.margin_end = 10;
+        content_area.margin_start = 10;
+        content_area.attach (title_label, 0, 0, 1, 1);
```
danrabbit: Since you're just attaching these widgets with the same width and all in one direction, you can use `.add` instead of `.attach`

PR #79, opened 2017-08-29
Show the real list of layouts

tintou:


src/Views/KeyboardLayoutView.vala
```diff
@@ -86,24 +87,139 @@ public class KeyboardLayoutView : AbstractInstallerView {
...
+            keyboard_layout_list_title.label = "<b>%s</b>".printf (((LayoutRow) row).layout.description);
+            var variants = ((LayoutRow) row).layout.variants;
+            keyboard_layout_list_box.add (new VariantRow ("", _("Default")));
+
+            foreach (var variant in variants.entries) {
+                keyboard_layout_list_box.add (new VariantRow (variant.key, variant.value));
+            }
+
             stack.visible_child = keyboard_layout_grid;
         });
```
davidmhewitt: I think you introduce at least one memory leak here. `doc` needs to be deleted when it is finished with. Code to load languages and variants can already be found here if you hadn't found it already: https://github.com/elementary/switchboard-plug-keyboard/blob/master/src/Layout/Handler.vala

tintou: Thanks for the memory leak, I didn't like the code on the plug, here we are only using one file. Let's hope GXml will be available on Debian soon so that we can forget libxml2…

davidmhewitt: The code in the plug only uses one file too? `evdev.xml` and `base.xml` are exactly the same.

I personally find the code in the plug to be more readable because of the XPaths, it's taking me a long time to understand how this XML parsing code works.

But it does work and the rest of the code looks OK, so I'll dismiss and approve so as not to block your progress.

PR #115, opened 2017-08-30
set up some back button stuff

danrabbit:


src/Views/AbstractInstallerView.vala
```diff
@@ -19,13 +19,16 @@
 public abstract class AbstractInstallerView : Gtk.Grid {
     public signal void cancel ();
     public bool cancellable { get; construct; }
+    public string previous_pane{ get; construct; }
```
donadigo: Space between `previous_pane` and `{`.

src/MainWindow.vala
```diff
@@ -59,6 +59,8 @@ public class Installer.MainWindow : Gtk.Dialog {
         stack.add_named (success_view, SUCCESS_VIEW);
         stack.add_named (error_view, ERROR_VIEW);
 
+        try_install_view.stack = stack;
```
donadigo: I would like to have a different solution than manually assiging the stack here, either the `AbstractInstallerView`  should have a `back ()` signal that should be handled within the `MainWindow` or you should explicitly pass the stack into the `AbstractInstallerView` class to avoid performing any operations on an `null` object.

PR #121, opened 2017-08-30
Tintou/layout

tintou:
Add libgnomekbd dependency and show the keyboard in a popover


davidmhewitt:
Is it correct that this is currently hardcoded to only show GB Colemak layout?

danrabbit:
No not really, but it's already not working on master, so we're just trying to avoid having conflicts

PR #160, opened 2017-09-01
Shutdown after 30s on success

danrabbit:
Do what we say we're gonna do


tintou:
Should the string be modified to reflect the remaining time ?

PR #171, opened 2017-09-01
Port the language logic to a new file, use data structures

tintou:


Appeltabak:
Forgot adding the new dep to the list in `README.md`

PR #182, opened 2017-09-03
Show the country selection the same way as the keyboard layout

tintou:


danrabbit:
It looks like for French the only option is `fr_CA`. I'm guessing that's not intentional ;)

tintou:
I'll fix this in another merge request ;)

PR #194, opened 2017-10-05
Don't rely on the partition table

tintou:


codygarver:
I can confirm this makes the installer not crash anymore, but code needs @jackpot51 review.

PR #198, opened 2017-10-06
Execute with pkexec

### LINKED ISSUE #195: Error opening /dev/sda: Permission denied
According to the log, running this command fails:
```
/dev/sda: Partitioning for Bios
parted -s --align optimal /dev/sda --mklabel msdos
Error opening /dev/sda: Permission denied
```

codygarver:
Fixes #195


codygarver:
If this policy to be translatable, I can make it so as a .xml.in, but I don't see any point since none of the text is meant to be user-facing.

PR #200, opened 2017-10-13
VariantWidget.vala: can't pass null to gtk.label

danrabbit:
Solves a terminal warning


src/Widgets/VariantWidget.vala
```diff
@@ -38,7 +38,7 @@ public class VariantWidget : Gtk.Frame {
         variant_scrolled.vexpand = true;
         variant_scrolled.add (variant_listbox);
 
-        back_button = new Gtk.Button.with_label (null);
```
tintou: Wouldn't it be better to just use `new Gtk.Button ();` ?

PR #206, opened 2017-12-12
Rename distinst config member from drive to disk

jackpot51:
This is done to match a change in distinst which means that the same terminology is now used in both the installer GUI and distinst backend: https://github.com/pop-os/distinst/blob/master/src/distinst.vapi#L17

This will require a new version of distinst, which will also give you automatic unmount and swapoff of any partitions the installer will attempt to utilize.

I believe the failure to swapoff may be the cause of installation failures experienced by @codygarver


jackpot51:
Closing in favor of #207

PR #214, opened 2018-01-24
Encryption screens

### LINKED ISSUE #213: Add a screen to select Encryption options
Distinst intends to support full-disk encryption so we need a screen to explain what that means and offer a choice of whether to do it or not

danrabbit:
Fixes #213 

This is blocked by:
- [x] Having the API available in distinst
- [x] Actually integrating with distinst. We'll want to pass a string if we're encrypting or `null` if we're not.
- [x] It also is currently in the wrong order
- [ ] ~Needs better icons~


cassidyjames:
The API is available in distinst. Just needs hooked up now.

cassidyjames:
@danrabbit to clarify, is integrating with encryption in distinst within the scope of this branch?

cassidyjames:
I'm fine with punting on better icons for later, the ones we have are fine for an mvp :stuck_out_tongue:

cassidyjames:
This builds with distinst from git, but the version in the repos (and thus CI) is out of date.

PR #216, opened 2018-01-25
Add flow chart to README

cassidyjames:


danrabbit:
I feel like this would be more appropriate for the wiki than the README.md

cassidyjames:
Agreed, gonna keep the flow chart image in the repo (since you can't upload directly to the wiki, it needs to be a known good URL for the wiki), but I'll remove it from the README.

PR #222, opened 2018-01-30
Update to new distinst API

jackpot51:
This updates to the new distinst API. This API requires more complex code to setup the disk, because custom partitioning is allowed with the API.

Do not merge just yet, I have work to do to cleanup, especially the error handling.


fabianthoma:
@jackpot51 protip for github. Reference the issues you close right in the commit messages or in the pull request. That will automatically close the issues when the Pull Request get's merged. 

This fixes #224

jackpot51:
Sorry about that @fabianthoma. That issue had not been created when I made the commits and the PR

jackpot51:
This PR is now ready to be merged.

src/Views/ProgressView.vala
```diff
@@ -16,6 +16,8 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+using Distinst;
```
cassidyjames: We tend to avoid `using` to make conflicting namespaces less ambiguous. Not sure if that'd be a blocker, but regardless I want to get someone more familiar with the Installer code to give this a look.

jackpot51: I removed it in https://github.com/elementary/installer/pull/222/commits/9826b6c3b43345bc76961faf336be9cae36acc4b

Issue #219, opened 2018-01-25
Move dialog layout views to two-pane layout


cassidyjames:
For consistency!

- [x] Issue detection #229 
- [x] Try or Install #235 
- [x] Drive Selection
- [x] Installation Success #232 
- [x] Installation Error #233 

Maybe more, add 'em if you find 'em

cassidyjames:
Super rough sketch for drives, but should communicate the idea. We should retain the same content, margins, and iconography, just re-laid-out to be side-by-side like the other views. The right side should be vertically center-aligned, but overflow into a scroll box (like how the current one works horizontally).

![drives-sketch](https://user-images.githubusercontent.com/611168/37610205-9b209dbe-2b64-11e8-88b7-c98ee2caf0d5.jpg)

cassidyjames:
Drive selection was merged in downstream at https://github.com/pop-os/installer/pull/19

Issue #205, opened 2017-11-29
Add flag for test mode


cassidyjames:
It would be super handy to be able to step through the installer in a fake or "test" mode that doesn't touch user data or do anything destructive, possibly with separate options for error and success results.

This would help with prototyping without having a local copy of distinst and/or help downstreams like Pop!_OS with theming.

cassidyjames:
We would probably also want to just set the progress view to have a timer so that it wouldn't just immediately skip that screen.

cassidyjames:
This should also log instead of actually restarting automatically

PR #232, opened 2018-02-22
Side by side success

cassidyjames:
Moves success view to side-by-side for #219.

Screenshot

Ready for review, but **don't merge as-is** since I'm overriding the default view and disabling the restart/poweroff for testing. I'm also not 100% sure about the copy changes, but the long title was super awkward. Open to suggestions.


src/Views/SuccessView.vala
```diff
@@ -29,29 +31,31 @@ public class SuccessView : AbstractInstallerView {
...
-        content_area.attach (secondary_label, 1, 1, 1, 1);
+
+        content_area.attach (image, 0, 0, 1, 1);
+        content_area.attach (title_label, 0, 1, 1, 1);
+        content_area.attach (description_label, 1, 0, 1, 2);
 
         var shutdown_button = new Gtk.Button.with_label (_("Shut Down"));
 
-        var restart_button = new Gtk.Button.with_label (_("Restart"));
+        var restart_button = new Gtk.Button.with_label (_("Restart Device"));
```
cassidyjames: More explicit about what's happening since we're relying less on the title copy. Restart the _device_, not the installation.

src/Views/SuccessView.vala
```diff
@@ -61,13 +65,15 @@ public class SuccessView : AbstractInstallerView {
...
+                // FIXME: Disabled while in development
+                // system_interface.power_off (false);
+                critical ("Fake power off");
             } catch (IOError e) {
                 critical (e.message);
             }
         });
 
-        Timeout.add_seconds (30, () => {
+        Timeout.add_seconds (RESTART_TIMEOUT, () => {
```
cassidyjames: Instead of magic numbers while I'm here

src/Views/SuccessView.vala
```diff
@@ -77,7 +83,9 @@ public class SuccessView : AbstractInstallerView {
 
     private void session_restart () {
         try {
-            system_interface.reboot (false);
+            // FIXME: Disabled while in development
```
cassidyjames: We should figure out a longer-term solution, like #205

src/Views/SuccessView.vala
```diff
@@ -29,29 +31,31 @@ public class SuccessView : AbstractInstallerView {
         var image = new Gtk.Image.from_icon_name ("process-completed", Gtk.IconSize.DIALOG);
         image.valign = Gtk.Align.START;
 
-        var primary_label = new Gtk.Label (_("Restart your device to continue setting up"));
-        primary_label.max_width_chars = 60;
-        primary_label.wrap = true;
-        primary_label.xalign = 0;
-        primary_label.get_style_context ().add_class ("h2");
+        var title_label = new Gtk.Label (_("Continue Setting Up"));
```
cassidyjames: Variable names changed for consistency with the other views.

cassidyjames:
Screenshot

cassidyjames:
@danrabbit mentioned in Slack that this looks a bit awkward since there's not much content on the right; the paragraph is just kind of floating there. Not sure how to address.

cassidyjames:
One thought I had was to pull out the "restarting" text into a primary label, but @danrabbit says it's awkward then because we have two main labels. (Ignore the FrankenJunoPop fonts and crappy alignment):

image

He wondered if we need to do the split layout at all for success and error; it was planned for Kate's designs, but we might still just be able to do that with CSS (replacing the dialog-style icon).

Issue #202, opened 2017-10-24
Explain that demo mode isn't optimal


danrabbit:
We should add some kind of copy to explain that while in demo mode performance may be impacted and some features may not be available (notably AppCenter)

PMiner:
demo mode being, booting from the live installation?

btkostner:
@PMiner yes

PMiner:
like a notification, a pop-up window, a little text in the corner? something like that?

lainsce:
Maybe add some floating text on top of the wallpaper in the bottom right or left with "Demo Mode - Some things might not work correctly" as maybe the copy?

SubhadeepJasu:
Having a sticky banner to the bottom corner which says "Demo mode isn't optimal" with two buttons "More Info" and "Install". Just an idea.

PiFi:
Maybe a sticky notification would fit with the OS theme better?

cassidyjames:
To clarify, the scope here is just within the Installer UI itself, not during the live session.

PiFi:
Yes

cassidyjames:
Would adding a fourth item in the row make sense, or do we start to run out of room?

i.e. from:

![elementary-try-or-install](https://user-images.githubusercontent.com/611168/33397880-4dc851de-d50a-11e7-9e7d-91ed48f05eb3.png)

to:

># Install elementary OS
>You can install elementary OS on this device now, or try Demo Mode without installing.
>
>:white_check_mark: | :arrow_left: | :arrow_down: | :no_entry_sign: 
>--- | --- | --- | ---
>Data from your previous operating system is unchanged in Demo Mode | Any changes you make in Demo Mode will not be saved | You can always return to the installer from Demo Mode by selecting the Install icon. | Performance and features in Demo Mode may not reflect the installed experience.

(terrible markdown mockup aside)

PiFi:
I think four might fit if you put them in a 2x2 grid

PiFi:
But that might look strange.

lainsce:
Maybe move the fourth item below the demo mode one, as a new paragraph, with a red icon to catch attention?

PiFi:
![33397880-4dc851de-d50a-11e7-9e7d-91ed48f05eb3](https://user-images.githubusercontent.com/20211219/33404006-4062385c-d528-11e7-8964-47a3e8451e64.png)

PiFi:
like this?

lainsce:
[...]

PR #235, opened 2018-02-22
Side by side Try or Install

cassidyjames:
For #219. Not 100% sold on the icons, but full-color icons felt too heavy there.

image

This design will also better accommodate #202

### DIFF
--- a/src/Views/TryInstallView.vala
+++ b/src/Views/TryInstallView.vala
@@ -35,10 +35,6 @@ public class TryInstallView : AbstractInstallerView {
         title_label.valign = Gtk.Align.START;
         title_label.get_style_context ().add_class ("h2");
 
-        // var description_label = new Gtk.Label (_("You can install %s on this device now, or try Demo Mode without installing.").printf (Utils.get_pretty_name ()));
-        // description_label.wrap = true;
-        // description_label.max_width_chars = 60;
-
         var choice_image = new Gtk.Image.from_icon_name ("computer-symbolic", Gtk.IconSize.LARGE_TOOLBAR);
         choice_image.halign = Gtk.Align.END;
         choice_image.valign = Gtk.Align.START;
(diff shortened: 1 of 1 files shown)


danrabbit:
I kinda feel like the main label here should be "Install or Try Demo mode". Since install is a verb, it feels weird to have the alternative be a noun.

It definitely feels weird that there's one row of just text with no icon. I wonder if this is either redundant or can be merged with another row or if maybe the two middle rows are redundant? Maybe that can be something like

"Changes you make in Demo Mode will not be saved. Data from your previous operating system will remain unchanged."

cassidyjames:
@danrabbit 

>I kinda feel like the main label here should be "Install or Try Demo mode". Since install is a verb, it feels weird to have the alternative be a noun.

Yeah, agreed.

>It definitely feels weird that there's one row of just text with no icon.

The text is in a different context though; we're explaining the choice here, then the tree points below are all specific to Demo Mode.

>I wonder if this is either redundant or can be merged with another row or if maybe the two middle rows are redundant?

I don't think the first label is redundant since it's explaining the choice. But the next two definitely could be combined, which then makes the three bullet points basically:

- Here's what the choice is
- Here's an implication of Demo Mode
- Here's how to get back here 

I like it.

src/Views/TryInstallView.vala
```diff
@@ -27,63 +27,65 @@ public class TryInstallView : AbstractInstallerView {
...
+        var return_label = new Gtk.Label (_("You can always return to the installer from Demo Mode by selecting the Install icon."));
+        return_label.halign = Gtk.Align.START;
+        return_label.max_width_chars = 52;
         return_label.valign = Gtk.Align.START;
         return_label.wrap = true;
-        return_label.set_justify (Gtk.Justification.CENTER);
+        return_label.xalign = 0;
+
+        var grid = new Gtk.Grid ();
+        grid.orientation = Gtk.Orientation.VERTICAL;
```
danrabbit: If you're just using `attach` and not `add` then setting orientation here doesn't really do anything

PR #241, opened 2018-03-07
Add test mode

### LINKED ISSUE #205: Add flag for test mode

cassidyjames:
Currently nothing listens to this, but it does add a label to identify it to the action area.

Adds a `--test` and `-t` CLI flag that sets `Installer.App.test_mode` to true.

Will fix #205


cassidyjames:
@donadigo suggested using the built-in `Granite.Application.DEBUG` flag for this, but I think we still want to be able to debug live installs (and let test mode be a separate thing).

PR #247, opened 2018-03-08
Travis: reliably use add-apt-repository

codygarver:


codygarver:
Sometimes `add-apt-repository` freaks out if you don't `apt-get update` after each use.

PR #262, opened 2018-03-19
Make the check view automatic

### LINKED ISSUE #250: "Connect to a Power Source" screen should disapear on ac power
When presented with the "Connect to a Power Source" screen, when a charger is plugged in, the installer should automatically detect it and continue on to the next screen.

### LINKED ISSUE #251: Consider power consumption warning
Currently if a machine is running off of battery power, a screen will appear recommending the user plug in to a power source.

_Something to consider_:
What if a user begins plugged in and then unplugs their machine after where the `Connect to a power source` screen appears.

Possible things to consider:

- A suggestion to plug in the machine towards the beginning of the installation
- Show the `Connect to a power source` screen at anytime before the actual installation begins

@cassidyjames thoughts?

mmstick:
Adds a timeout function which monitors requirements and switches to the check view when
requirements are not met. Clicking Ignore will ignore future requirement checks. This
will allow the user to simply plug their laptop in to get the view to automatically
switch back to the view that they were on.

Closes #250, closes #251

### DIFF
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -82,25 +82,35 @@ public class Installer.MainWindow : Gtk.Dialog {
         try_install_view.previous_view = keyboard_layout_view;
         stack.add (try_install_view);
         stack.visible_child = try_install_view;
+        load_checkview ();
 
-        try_install_view.next_step.connect (() => load_checkview ());
+        try_install_view.next_step.connect (() => load_encryptview ());
     }
 
     private void load_checkview () {
-        if (check_view != null) {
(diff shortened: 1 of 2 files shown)


tintou:
We don't need a timeout here, we can get a signal trigger from DBus when the computer is plugged again

cassidyjames:
I really like this UX-wise, as it makes the check view more useful in other contexts (like we mentioned in both #250 and #251).

If merged, this fixes #251 as well.

mmstick:
@tintou Have any examples on how to do that?

PR #263, opened 2018-03-19
Distinst Drive Detection

### LINKED ISSUE #180: Drive View: Filter drives by estimated install size
Gray out or don't show drives that are below a sane threshold.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/48903737-drive-view-filter-drives-by-estimated-install-size?utm_campaign=plugin&utm_content=tracker%2F46189108&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F46189108&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #209: distinst crashes immediately without hardcoding enough_space = true
I can't install on my machine, possibly because there's a 4MB partition on the disk. Even though there are other disks/partitions with plenty of space.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/53022401-distinst-crashes-immediately-without-hardcoding-enough_space-true?utm_campaign=plugin&utm_content=tracker%2F46189108&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F46189108&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #230: Filter out boot drive
Do not show the drive you are installing from as an option in the install menu

### LINKED ISSUE #231: Use distinst for drive detection
Instead of UDisks2, use distinst for drive detection. This will ensure that all information distinst provides is available before the progressview

mmstick:
Closes #180, Closes #209, Closes #230, Closes #231


cassidyjames:
With this change I'm getting a much more verbose and seemingly machine-friendly string for the drive names, instead of the more human-friendly names like before.

cassidyjames:
Before:
image

After:
image

mmstick:
@cassidyjames How's this?

screenshot from 2018-03-19 16-51-51

mmstick:
screenshot from 2018-03-19 18-06-33

Now checking for disk sizes

src/Views/DiskView.vala
```diff
@@ -142,21 +144,26 @@ public class Installer.DiskView : AbstractInstallerView {
...
-                    disk_grid.get_children ().foreach ((child) => {
-                        ((Gtk.ToggleButton)child).active = child == disk_button;
-                    });
 
-                    next_button.sensitive = true;
-                } else {
-                    next_button.sensitive = false;
-                }
-            });
+            if (size < 5000000000) {
```
jackpot51: This should probably be a configuration option, or a value read from the `/cdrom/casper/filesystem.size` file

mmstick: @jackpot51 Think I should implement that logic in distinst?

cassidyjames:
Is `/sdb` and `/sdc` really useful to people in this context? At this point, the user is not on their familiar/installed system, so these paths might as well be meaningless. I would prefer we didn't diverge from the initial design without discussion and good reason; we've considered a lot of these things when we spent the time actually designing it.

cassidyjames:
Invalid drives should be sorted to the end, as [covered in the Wiki](https://github.com/elementary/installer/wiki/Drive-Selection).

jackpot51:
@cassidyjames I agree with sorting the invalids at the end.

The /dev/sdb is useful for cross-referencing the drive in other tools like GParted.

jackpot51:
Especially if you have two identical drives as is common in many systems. For example, my desktop at home has two Samsung EVO 850 SSDs. Seing simply "Samsung EVO 850" twice does not give me enough information to decide which drive to install to.

Perhaps when we have OS detection, this will be easier to determine with an OS icon and/or name, but for now I think it is very useful to be able to see the path that other tools will display when referring to the drive.

cassidyjames:
@jackpot51 the custom partition view is a separate concern, as that will be an entirely different view with different information. If you're using this screen, it's not likely that you're going to know or care what the path to the drive is. And if the intent is to distinguish two identical drives, we could be smart about actually solving that by only showing it when the drives are in fact the same.

Is the path likely to be the same across OSes/boots? If so, I guess I'm fine with showing it, but it need to be secondary information instead of on the same line as the drive name.

jackpot51:
installer-sort

I addressed those two items - moving the path to the next line and putting disabled disks at the end.

isantop:
I do agree with @jackpot51 here. Displaying two completely identical entries with no way to distinguish them seems like a bad experience, especially when one has my important daily-driver OS that needs to be stable, while the other one is my willy-nilly test drive for whatever I feel like running. Even if the drive letters are different from my normal OS (They're very likely to be the same among all Ubuntu-based OSs from the last decade or so), displaying the /dev path allows me to use other tools within the system if necessary to separate them and install the OS to the correct drive, preventing me from losing important data.

PR #264, opened 2018-03-20
Update to new distinst keyboard config API

mmstick:
Distinst now takes an optional `model` and `variant`, in addition to a `layout`.


jackpot51:
@codygarver We will need to update distinst to the 785827d version, I will let you know when that is in our PPA

PR #265, opened 2018-03-20
Check for Root

### LINKED ISSUE #196: Check for root permissions
Installation fails without root permissions so Installer shouldn't proceed without them.

mmstick:
If the program is not being run by root, the following view will be displayed (using a CheckView view).

Closes #196

screenshot from 2018-03-20 14-25-11


cassidyjames:
I think throwing a PolicyKit authentication is more appropriate than just hard stopping with the installation, but I'm not sure how much of a change that would be. Assigning @tintou for input.

danrabbit:
Yeah I agree with Cass. It would be smarter to throw a polkit dialog and auth in

tintou:
Yeah and things will get much more complicated with wayland…
I'm unsure that it is wanted to have a newbie-proof installer, if they use the right ISO everything should be fine.

tintou:
So the code is good, but I'm :-1: on this feature. Don't add unnecessary complexity.

PR #266, opened 2018-03-20
session: remove hardcoded GSD components

codygarver:


codygarver:
These were added to support the GSD version between Loki and Bionic, but now that 3.27 has landed in Bionic, not only is this unnecessary since these all have autostarts in /etc/xdg/autostart, it breaks the session entirely, it won't even start.

PR #268, opened 2018-03-21
Fix double free of Disks pointer

mmstick:
Fixes an issue that causes the installer to crash after the `Installer.install()` method is used, due to `install()` taking ownership of `disks`, then Vala also trying to free it.


davidmhewitt:
This makes sense in combination with https://github.com/pop-os/distinst/commit/771417a0471e2aeaa10bad83f38b06e9cc5483a8 , merging.

Issue #208, opened 2017-12-22
country and keyboard values in Configuration are not set


jackpot51:
After completing the language and keyboard selection, the Configuration object has null for the country and keyboard

Issue #89, opened 2017-08-29
Keyboard View: Automatically highlight/scroll to language


danrabbit:
Based on the selection of the language from the previous screen

PR #269, opened 2018-03-22
Use select button to move to region or variant selection

jackpot51:
This is split into 3 completely independent commits, that can potentially be cherry picked, or merged in entirety:

## Force a country to be selected

This fixes https://github.com/elementary/installer/issues/208 by requiring that a country is selected before proceeding.

Commit: https://github.com/elementary/installer/commit/ea978cff6c326eccf2135877bf40dde327f9b761

## Select keyboard layout based on region

This fixes https://github.com/elementary/installer/issues/89 as best as possible, by looking for a layout with the same name as the country code. For example `US` country code => `us` layout. I need help figuring out the right way to select the row, currently I am using `Idle.add` and running a callback there.

Commit: https://github.com/elementary/installer/commit/3dba678f5aa4574473e5269b801980baa0a87845

## Use select button to move to region or variant selection

I believe this improves the movement through the installer, but it is completely a UX change and does not change functionality. I see it potentially being desired to use the back button in a similar way, but that would also be more controversial and would need discussion.

Commit: https://github.com/elementary/installer/commit/0b0fd1de979cc0c986e16f57fa53b1199a8bf33d


src/Views/KeyboardLayoutView.vala
```diff
@@ -78,22 +78,34 @@ public class KeyboardLayoutView : AbstractInstallerView {
 
         back_button.clicked.connect (() => ((Gtk.Stack) get_parent ()).visible_child = previous_view);
 
-        next_button.clicked.connect (() => next_step ());
+        next_button.clicked.connect (() => {
+            unowned Gtk.ListBoxRow row = input_variant_widget.main_listbox.get_selected_row ();
+            if (row != null) {
+                var layout = ((LayoutRow) row).layout;
+                Configuration.get_default ().keyboard_layout = layout.name;
```
donadigo: Since you're using `Configuration.get_default ()` multiple times here, I suggest getting the instance once: 
`unowned Configuration = Configuration.get_default ()`.

src/Views/KeyboardLayoutView.vala
```diff
@@ -78,22 +78,34 @@ public class KeyboardLayoutView : AbstractInstallerView {
...
+
+                unowned Gtk.ListBoxRow vrow = input_variant_widget.variant_listbox.get_selected_row ();
+                if (vrow != null) {
+                    string variant = ((VariantRow) vrow).code;
+                    Configuration.get_default ().keyboard_variant = variant;
+                } else if (layout.variants.is_empty) {
+                    Configuration.get_default ().keyboard_variant = null;
+                } else {
+                    row.activate();
```
donadigo: Space between `(`.

src/Views/KeyboardLayoutView.vala
```diff
@@ -78,22 +78,34 @@ public class KeyboardLayoutView : AbstractInstallerView {
...
+                    return;
+                }
+            } else {
+                error ("next_button enabled when no keyboard selected");
```
donadigo: You should not use `error ()` in GUI applications, as this function will just abort the process and make the window close, even if it's really unlikely that this if branch gets executed.

src/Views/LanguageView.vala
```diff
@@ -118,15 +119,24 @@ public class Installer.LanguageView : AbstractInstallerView {
...
-                string country = ((CountryRow) crow).country_entry.alpha_2;
-                Configuration.get_default ().country = country;
+                unowned Gtk.ListBoxRow crow = lang_variant_widget.variant_listbox.get_selected_row ();
+                if (crow != null) {
+                    string country = ((CountryRow) crow).country_entry.alpha_2;
+                    Configuration.get_default ().country = country;
+                } else if (lang_entry.countries.length == 0) {
+                    Configuration.get_default ().country = null;
+                } else {
+                    row.activate();
```
donadigo: Space before `(`.

src/Views/LanguageView.vala
```diff
@@ -118,15 +119,24 @@ public class Installer.LanguageView : AbstractInstallerView {
...
+                    return;
+                }
+            } else {
+                error ("next_button enabled when no language selected");
```
donadigo: Same here with `error ()`.

src/Views/KeyboardLayoutView.vala
mmstick: @donadigo Does Vala have a code formatting tool? That would make this a lot easier.

donadigo: @mmstick I think Uncrustify supports Vala natively, we also have a Vala linter in the works to be able to point out issues like that: https://github.com/elementary/vala-lint

jackpot51:
@donadigo all of those comments have been addressed.

src/Views/KeyboardLayoutView.vala
```diff
@@ -142,6 +149,24 @@ public class KeyboardLayoutView : AbstractInstallerView {
         }
 
         show_all ();
+
+        Idle.add (() => {
```
donadigo: It looks like this Idle clause is getting really big for something that is a lambda function, I suggest putting the contents of that to another function that will get called by `Idle.add`.

src/Views/KeyboardLayoutView.vala
```diff
@@ -78,22 +78,36 @@ public class KeyboardLayoutView : AbstractInstallerView {
 
         back_button.clicked.connect (() => ((Gtk.Stack) get_parent ()).visible_child = previous_view);
 
-        next_button.clicked.connect (() => next_step ());
+        next_button.clicked.connect (() => {
```
donadigo: Could you put the contents of this into another function?

cassidyjames:
Currently we have two different navigation schemes: between views, and within views. For language and keyboard, we intended to keep the within-view navigation separate, using the row selection and the inline back button. However, with the encryption view we are mixing the two navigation schemes.

I agree that we should clean this up somehow, but I think it's outside the scope of the other changes here. I'd revert https://github.com/elementary/installer/commit/0b0fd1de979cc0c986e16f57fa53b1199a8bf33d for now and I'll file UX issues to clean up the navigation.

I'm open to input from other @elementary/ux peeps as well.

PR #271, opened 2018-03-22
Fix check view navigation

jackpot51:
Before, the check view would be launched before consenting to the install and would then go immediately to the encrypt view if no error was found, not showing the user the demo mode option.

With this patch, the check view will be shown after the user has a chance to select demo mode and before encryption.

### DIFF
--- a/src/Views/AbstractInstallerView.vala
+++ b/src/Views/AbstractInstallerView.vala
@@ -19,6 +19,8 @@
 public abstract class AbstractInstallerView : Gtk.Grid {
     public bool cancellable { get; construct; }
     public unowned Gtk.Widget? previous_view { get; set; }
+    
+    public signal void previous_step ();
 
     protected Gtk.Grid content_area;
     protected Gtk.ButtonBox action_area;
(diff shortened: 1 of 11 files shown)


tintou:
The problem is that this view can be shown at any time after the try_install_view, this change forces the user to go back to the encryption view…

jackpot51:
I see. I can fix that. Do you agree with this statement?

Before, the check view would be launched before consenting to the install and would then go immediately to the encrypt view if no error was found, not showing the user the demo mode option.

jackpot51:
There are all kinds of issues with the way it currently is, I will try to fix them all:

- Clicking Select on the keyboard layout screen skips the Try or Install screen
- Clicking Cancel Installation takes you back to the previous screen, rather than the Try or Install screen
- Going back to language selection, then plugging power back in will take you to the try or install screen

jackpot51:
All three issues I found are now fixed with the latest commit.

jackpot51:
- `previous_view` indicates which view the check view should return to when "Cancel Installation" is clicked
- `ignore_view` indicates which view to return to when "Ignore" is clicked
- If the "Ignore" button is clicked or the requirements are met, it will return to the `ignore_view`
- If the "Cancel Installation" button is clicked, it must return to the Install or Try view, and clear the `ignore_view` so that meeting the requirements will not move the installer forward in steps

src/Views/AbstractInstallerView.vala
```diff
@@ -46,7 +48,7 @@ public abstract class AbstractInstallerView : Gtk.Grid {
         if (cancellable) {
             var cancel_button = new Gtk.Button.with_label (_("Cancel Installation"));
             cancel_button.clicked.connect (() => {
-                ((Gtk.Stack) get_parent ()).visible_child = previous_view;
```
jackpot51: Has this line been moved elsewhere? It will need to be called for the other views.

PR #273, opened 2018-03-23
Implement Better Terminal Behavior

### LINKED ISSUE #227: Terminal output scroll
Current experience:
When opening the terminal output during the installation, the window always shows the top/beginning of the output. Manual scrolling is required to see the latest progress of the installation.

Expected experience:
When selecting the terminal button, the output should automatically scroll to the bottom to show the most recent output. Whenever this screen is scrolled all the way down, it should continue to scroll when new input is added. If the user scrolls up, automatic scrolling should stop.

mmstick:
Closes #227

- Makes a change to how the log buffer is handled. Concatenating text to the buffer directly causes the text view to reset, losing the scrolled position. Instead, using the `insert` method on the buffer allows the existing text contents to be retained, and new contents inserted at the end.
- The terminal button will now automatically scroll to the bottom of the view when the inner text view is updated, and the view was previously at the bottom.

### DIFF
--- a/src/Views/ProgressView.vala
+++ b/src/Views/ProgressView.vala
@@ -20,7 +20,6 @@ public class ProgressView : AbstractInstallerView {
     public signal void on_success ();
     public signal void on_error ();
 
-    private bool overshot;
     private double prev_upper_adj = 0;
     private Gtk.ScrolledWindow terminal_output;
     public Gtk.TextView terminal_view { get; construct; }
--- a/src/Views/ProgressView.vala
+++ b/src/Views/ProgressView.vala
@@ -20,7 +20,8 @@ public class ProgressView : AbstractInstallerView {
     public signal void on_success ();
     public signal void on_error ();
 
-    private double last_upper;
+    private bool overshot;
+    private double prev_upper_adj = 0;
     private Gtk.ScrolledWindow terminal_output;
     public Gtk.TextView terminal_view { get; construct; }
     private Gtk.ProgressBar progressbar;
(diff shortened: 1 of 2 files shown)


src/Views/ProgressView.vala
```diff
@@ -74,14 +75,31 @@ public class ProgressView : AbstractInstallerView {
...
+                scroll_to_bottom ();
+            }
+        });
+
         show_all ();
     }
 
+    private void scroll_to_bottom () {
+        var adj = terminal_output.vadjustment;
+        adj.value = adj.upper;
```
jackpot51: Should this be:
```
adj.value = adj.upper - adj.page_size;
```

jackpot51:
This is not quite working for me, and I am not sure why. I opened the terminal immediately after getting to the progress view, and it scrolled for about two lines before no longer scrolling.

src/Views/ProgressView.vala
jackpot51: I don't think the change I proposed works to fix the issues I have seen either

mmstick:
I may need to account for the height of the last line(s) added, in case there's more than one.

jackpot51:
Yep, that sounds right

mmstick:
After much mangling with values, I think I've got it. Need to test a bit more, though.

jackpot51:
Sounds good @mmstick. I will take a look when it is committed

jackpot51:
That looks better @mmstick

jackpot51:
There is a point early in the install process where distinst outputs more than one page of content, and it does track the bottom of the output

jackpot51:
If you add a GLib.Thread.usleep(1000000) to the start_installation function, you will have time to open the log view before any output

mmstick:
Going to try a different approach.

mmstick:
It took countless hours, and multiple attempts with different approaches, but the new solution should work better. Modified distinst locally to generate all sorts of random log data of varying numbers of newlines, at random intervals (from bursts multiple pages long to once a second).

The only thing I haven't figured out is how to prevent the ScrolledWindow from rubber banding back when the bottom end is overshot. Using the scrollbar to scroll to the end doesn't cause that behavior. Couldn't find any documentation from GTK or elsewhere on the matter.

PR #275, opened 2018-03-23
Linter

btkostner:
Adds `vala-lint` docker image to CI tests


jackpot51:
This is a good idea, and the new .travis.yml looks good

PR #277, opened 2018-03-26
Force region to be selected, set layout based on region

jackpot51:
A simpler version of https://github.com/elementary/installer/pull/269 that does not change navigation


jackpot51:
Thanks @cassidyjames

PR #284, opened 2018-04-02
[WIP] Custom Partitioning View

mmstick:
This will need some insights from the Vala experts, and @cassidyjames regarding the UI. I have been able to use the new partitioning view to perform an installation, so everything should be working (minus getting the partition boxes to use their specified widths...).

[screenshot from 2018-04-02 16-13-37](https://www.youtube.com/watch?v=EIcOGg5jRHg)

(click on image for video)

## Todo

- [ ] Integrate GParted / Use existing distinst functions to do partitioning
- [ ] Support configuring logical devices and encrypted installations

### DIFF
--- a/data/application.css
+++ b/data/application.css
@@ -22,5 +22,51 @@ textview.terminal {
 
 textview.terminal selection {
     background-color: #93a1a1;
-    color: #252e32; 
+    color: #252e32;
+}
+
+.partition {
+    border-style: solid;
+    border-right-width: 1px;
+}
+
+.disk-bar {
+    border-style: solid;
+    border-width: 1px;
+}
+
+.ext2, .ext3, .ext4 {
+    background-color: rgb(114, 147, 203);
+}
+
+.fat16, .fat32 {
+    background-color: rgb(225, 151, 76);
+}
+
+.btrfs {
+  background-color: rgb(132, 186, 91);
+}
+
+.xfs {
+  background-color: rgb(211, 94, 96);
+}
+
+.ntfs {
+  background-color: rgb(144, 103, 167);
+}
+
+.luks {
+  background-color: rgb(171, 104, 87);
+}
+
+.lvm {
+  background-color: rgb(204, 194, 16);
+}
+
+.none {
+  background-color: rgb(128, 133, 133);
+}
+
+.swap {
+  background-color: rgb(218, 124, 48);
 }
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -26,6 +26,7 @@ public class Installer.MainWindow : Gtk.Dialog {
     private TryInstallView try_install_view;
     private Installer.CheckView check_view;
     private DiskView disk_view;
+    private PartitioningView partition_view;
     private ProgressView progress_view;
     private SuccessView success_view;
     private EncryptView encrypt_view;
@@ -88,6 +89,7 @@ public class Installer.MainWindow : Gtk.Dialog {
         stack.add (try_install_view);
         stack.visible_child = try_install_view;
 
+        try_install_view.custom_step.connect (() => load_partitioning_view ());
         try_install_view.next_step.connect (() => load_encryptview ());
     }
 
@@ -129,6 +131,23 @@ public class Installer.MainWindow : Gtk.Dialog {
         set_check_view_visible (!check_ignored && !check_view.check_requirements ());
     }
 
+    private void load_partitioning_view() {
+        if (partition_view != null) {
+            partition_view.destroy ();
+        }
+
+        partition_view = new PartitioningView (minimum_disk_size);
+        partition_view.previous_view = try_install_view;
+        stack.add (partition_view);
+        stack.visible_child = partition_view;
+
+        load_checkview ();
+
+        partition_view.next_step.connect (() => {
+            load_progress_view ((owned) partition_view.mounts);
+        });
+    }
+
     private void load_encryptview () {
         if (encrypt_view != null) {
             encrypt_view.destroy ();
@@ -159,15 +178,15 @@ public class Installer.MainWindow : Gtk.Dialog {
             stack.visible_child = try_install_view;
         });
 
-        disk_view.next_step.connect (() => load_progress_view ());
+        disk_view.next_step.connect (() => load_progress_view (null));
     }
 
-    private void load_progress_view () {
+    private void load_progress_view (GLib.Array<Installer.Mount>? disk_config) {
         if (progress_view != null) {
             progress_view.destroy ();
         }
 
-        progress_view = new ProgressView ();
+        progress_view = new ProgressView (disk_config);
         stack.add (progress_view);
         stack.visible_child = progress_view;
 
--- a/src/Objects/Mount.vala
+++ b/src/Objects/Mount.vala
@@ -34,8 +34,14 @@ public class Installer.Mount {
         this.format = format;
     }
 
-    public bool has_esp_fs () {
+    public bool is_valid_boot_mount () {
         return filesystem == Distinst.FileSystemType.FAT16
             || filesystem == Distinst.FileSystemType.FAT32;
+    }
+
+    public bool is_valid_root_mount () {
+        return filesystem != Distinst.FileSystemType.FAT16
+            && filesystem != Distinst.FileSystemType.FAT32
+            && filesystem != Distinst.FileSystemType.NTFS;
     }
 }
--- a/src/Objects/Mount.vala
+++ b/src/Objects/Mount.vala
@@ -1,0 +1,41 @@
+// -*- Mode: vala; indent-tabs-mode: nil; tab-width: 4 -*-
+/*-
+ * Copyright (c) 2016-2018 elementary LLC. (https://elementary.io)
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * Authored by: Michael Aaron Murphy <michael@system76.com>
+ */
+
+public class Installer.Mount {
+    public string partition_path;
+    public string parent_disk;
+    public string mount_point;
+    public Distinst.FileSystemType filesystem;
+    public bool format;
+
+    public Mount (string partition, string parent_disk, string mount,
+                  bool format, Distinst.FileSystemType fs) {
+        partition_path = partition;
+        this.parent_disk = parent_disk;
+        mount_point = mount;
+        filesystem = fs;
+        this.format = format;
+    }
+
+    public bool has_esp_fs () {
+        return filesystem == Distinst.FileSystemType.FAT16
+            || filesystem == Distinst.FileSystemType.FAT32;
+    }
+}
--- a/src/Views/PartitioningView.vala
+++ b/src/Views/PartitioningView.vala
@@ -109,9 +109,9 @@ public class Installer.PartitioningView : AbstractInstallerView  {
         for (int i = 0; i < mounts.length; i++) {
             var m = mounts.index (i);
 
-            if (m.mount_point == "/") {
+            if (m.mount_point == "/" && m.is_valid_root_mount ()) {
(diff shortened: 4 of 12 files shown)


mmstick:
Closing as it has been merged in the pop-os fork.

cassidyjames:
Re-opening for critique and review

mmstick:
Improvements have been made to it since this was closed, so critiques may or may not remain valid.

src/Views/TryInstallView.vala
```diff
@@ -90,6 +91,8 @@ public class TryInstallView : AbstractInstallerView {
 
         var demo_button = new Gtk.Button.with_label (_("Try Demo Mode"));
 
+        var custom_button = new Gtk.Button.with_label (_("Custom Partitioning"));
```
cassidyjames: wtaf

The "Custom Partitioning" button here should read `Customize Partitions…` (with the ellipsis since it takes the user elsewhere without performing an action on the current view) and be in [the incidental button position](https://elementary.io/docs/human-interface-guidelines#dialogs) on the DiskView.

src/Views/PartitioningView.vala
```diff
@@ -0,0 +1,158 @@
...
+            }
+
+            var disk_bar = new DiskBar (model, path, size, partitions);
+            disk_list.attach(disk_bar.label, 0, id, 1, 1);
+            disk_list.attach(disk_bar, 1, id, 1, 1);
+
+            id += 1;
+        }
+
+        next_button = new Gtk.Button.with_label (_("Erase and Install"));
```
cassidyjames: image

Something weird is happening with the buttons here, it's like they weren't packed inside the correct action_area.

src/Widgets/DiskBar.vala
```diff
@@ -0,0 +1,85 @@
...
+        this.size = size;
+
+        generate_label ();
+        generate_bar (partitions);
+    }
+
+    construct {
+        this.orientation = Gtk.Orientation.HORIZONTAL;
+        this.hexpand = true;
+        this.get_style_context ().add_class("disk-bar");
```
cassidyjames: I think instead of recreating this look with custom classes and styling, we could lean a little more on the existing `.storage-bar` styles that are already implemented in the elementary and Pop! stylesheets. The container box would need `.storage-bar`, the disk would need `.trough`, and then each partition inside would need `.fill-block` and then could have a custom class for coloring if we want.

This would set borders, shadows, padding, etc. to what the stylesheet already uses to represent a disk and its contents elsewhere.

data/application.css
```diff
@@ -22,5 +22,51 @@ textview.terminal {
...
+    border-right-width: 1px;
+}
+
+.disk-bar {
+    border-style: solid;
+    border-width: 1px;
+}
+
+.ext2, .ext3, .ext4 {
+    background-color: rgb(114, 147, 203);
```
cassidyjames: Where are these colors coming from? They do not pass any sort of contrast check. We should be sure to set all light colors where our foreground is more likely to contrast, or also be sure to set the foreground color to something that contrasts better.

src/Widgets/DiskBar.vala
```diff
@@ -0,0 +1,85 @@
...
+
+    private void generate_label () {
+        var name_label = new Gtk.Label (disk_name);
+        var size_label = new Gtk.Label ("<small>%s %s</small>".printf (disk_path, GLib.format_size (size)));
+        size_label.use_markup = true;
+        size_label.get_style_context ().add_class (Gtk.STYLE_CLASS_DIM_LABEL);
+
+        label = new Gtk.Box (Gtk.Orientation.VERTICAL, 6);
+        label.pack_start (name_label, false, false, 0);
+        label.pack_start (size_label, false, false, 0);
```
cassidyjames: Both of these labels should be halign `END`

cassidyjames:
Here is a quick inspector-mockup of using Granite.StorageBar classes for the labels and blocks. This also allows us to scale the partition by its relative size, not have to worry about label contrast, plus show free space.

image

src/Widgets/DiskBar.vala
mmstick: By container box, are you referring to the  disk bar + disk label?

data/application.css
mmstick: Colors were selected at random

danrabbit: Can we use colors from the palette? :) https://elementary.io/docs/human-interface-guidelines#color

src/MainWindow.vala
```diff
@@ -129,6 +131,23 @@ public class Installer.MainWindow : Gtk.Dialog {
         set_check_view_visible (!check_ignored && !check_view.check_requirements ());
     }
 
+    private void load_partitioning_view() {
```
danrabbit: Space before ()

src/Objects/Mount.vala
```diff
@@ -0,0 +1,47 @@
...
+    public Distinst.FileSystemType filesystem;
+    public bool format;
+
+    public Mount (string partition, string parent_disk, string mount,
+                  bool format, Distinst.FileSystemType fs) {
+        partition_path = partition;
+        this.parent_disk = parent_disk;
+        mount_point = mount;
+        filesystem = fs;
+        this.format = format;
```
danrabbit: New files should use GObject style. So that means that in the constructor should be something like:

```
Object (
    partition_path: partition_path,
    parent_disk: parent_disk,
    [...]
);
```

src/Objects/Mount.vala
```diff
@@ -0,0 +1,47 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * Authored by: Michael Aaron Murphy <michael@system76.com>
+ */
+
+public class Installer.Mount {
+    public string partition_path;
+    public string parent_disk;
```
danrabbit: If this (and others) is a construct property then it should be `public string parent_disk { get; construct; }

src/Views/PartitioningView.vala
```diff
@@ -0,0 +1,158 @@
...
+    private Gtk.Button next_button;
+    private Distinst.Disks disks;
+    private GLib.Array<PartitionBar> partitions;
+
+    public GLib.Array<Installer.Mount> mounts;
+
+    public static uint64 minimum_disk_size;
+
+    public PartitioningView (uint64 size) {
+        minimum_disk_size = size;
```
danrabbit: This should be part of your `Object` as in:

```
Object (
    cancellable: true,
    minimum_disk_size: size
);
```

src/Views/PartitioningView.vala
```diff
@@ -0,0 +1,158 @@
...
+        var disk_list = new Gtk.Grid ();
+        disk_list.row_spacing = 24;
+        var disk_scroller = new Gtk.ScrolledWindow (null, null);
+        disk_scroller.expand = true;
+        disk_scroller.vscrollbar_policy = Gtk.PolicyType.NEVER;
+        disk_scroller.add (disk_list);
+
+        var description = new Gtk.Label (_("Select which partitions to use across all drives. This will erase all data on the selected partitions."));
+
+        this.content_area.attach(description, 0, 0, 1, 1);
```
danrabbit: It's unnecessary to use `this`.
Space before ()
Default width and height for grid is `1` so you can omit the last two args here

src/Views/PartitioningView.vala
```diff
@@ -0,0 +1,158 @@
...
+                var part = partitions.index(i);
+                part.add_events (Gdk.EventMask.BUTTON_PRESS_MASK);
+                part.button_press_event.connect (() => {
+                    part.show_popover();
+                });
+            }
+
+            var disk_bar = new DiskBar (model, path, size, partitions);
+            disk_list.attach(disk_bar.label, 0, id, 1, 1);
+            disk_list.attach(disk_bar, 1, id, 1, 1);
```
danrabbit: space before `()`

src/Views/ProgressView.vala
```diff
@@ -20,13 +20,22 @@ public class ProgressView : AbstractInstallerView {
...
+
     private double prev_upper_adj = 0;
     private Gtk.ScrolledWindow terminal_output;
     public Gtk.TextView terminal_view { get; construct; }
     private Gtk.ProgressBar progressbar;
     private Gtk.Label progressbar_label;
     private const int NUM_STEP = 5;
 
+    public ProgressView (GLib.Array<Installer.Mount>? mounts) {
+        if (mounts == null) {
```
danrabbit: Per GObject style, this should go in the `construct block`

src/Views/ProgressView.vala
```diff
@@ -20,13 +20,22 @@ public class ProgressView : AbstractInstallerView {
...
+            stderr.printf ("mounts is null\n");
+        }
+        this.disk_config = mounts;
```
danrabbit: This should be in an `Object (disk_config: mounts);`

src/Widgets/DiskBar.vala
```diff
@@ -0,0 +1,85 @@
...
+
+    public Gtk.Box label;
+
+    public DiskBar (
+        string model,
+        string path,
+        uint64 size,
+        GLib.Array<PartitionBar> partitions
+    ) {
+        this.disk_name = model;
```
danrabbit: GObject style please :)

src/Widgets/PartitionBar.vala
```diff
@@ -0,0 +1,88 @@
...
+
+    public uint64 start;
+    public uint64 end;
+    public uint64 used;
+
+    public Distinst.Partition* info;
+    public Gtk.Label label;
+    public PartitionMenu menu;
+
+    public PartitionBar(Distinst.Partition* part, string parent_path,
```
danrabbit: Space before ()

src/Widgets/PartitionMenu.vala
```diff
@@ -0,0 +1,193 @@
...
+    private Gtk.Grid grid;
+    private Gtk.Switch use_partition;
+    private Gtk.ComboBoxText use_as;
+    private Gtk.Entry custom;
+    private Gtk.ComboBoxText type;
+    private string partition_path;
+    private string parent_disk;
+
+    public PartitionMenu (string path, string parent, SetMount set_mount, UnsetMount unset_mount) {
+        partition_path = path;
```
danrabbit: GObject style please :)

src/Widgets/PartitionMenu.vala
```diff
@@ -0,0 +1,193 @@
...
+        parent_disk = parent;
+        grid = new Gtk.Grid ();
```
danrabbit: Let's keep all the grid properties and methods together, so this should go down in front of line 52

src/Widgets/PartitionBar.vala
```diff
@@ -0,0 +1,88 @@
...
+        container = new Gtk.Box (Gtk.Orientation.VERTICAL, 0);
+        container.pack_start (label, true, true, 0);
+        var context = container.get_style_context ();
+        context.add_class (filesystem);
+        context.add_class ("partition");
+
+        menu = new PartitionMenu (path, parent_path, set_mount, unset_mount);
+        menu.relative_to = container;
+
+        add(container);
```
cassidyjames: Space before `(`

src/Widgets/PartitionMenu.vala
```diff
@@ -0,0 +1,193 @@
...
+        type.append_text ("fat32");
+        type.append_text ("btrfs");
+        type.append_text ("xfs");
+        type.append_text ("ntfs");
+        type.set_active (0);
+
+        grid.attach(use_partition, 1, 0, 1, 1);
+        grid.attach(use_as, 1, 1, 1, 1);
+        grid.attach(custom, 1, 2, 1, 1);
+        grid.attach(type, 1, 3, 1, 1);
```
cassidyjames: Space before `(`

src/Widgets/PartitionMenu.vala
```diff
@@ -0,0 +1,193 @@
...
+            } else if (active == 3) {
+                type_label.set_visible (false);
+                type.set_visible (false);
+            } else {
+                type_label.set_visible (true);
+                type.set_visible (true);
+                type.set_sensitive (true);
+            }
+        });
+        type.changed.connect(() => check_values (set_mount));
```
cassidyjames: Space before `(`

src/Widgets/DiskBar.vala
mmstick: I don't quite get the point of the `Object` thing in Vala. Some input parameters are incompatible with it, so you end up having an odd mix of Object + non-Object initialization.

jackpot51:
This will be resubmitted at a later time, from a branch of the pop-os fork. There are several things that need to be merged beforehand, such as encryption support for BIOS, which I will submit a PR for, before the newest version of this will rebase correctly on elementary/installer.

src/Objects/Mount.vala
mmstick: These are both construct and non-construct fields.

PR #294, opened 2018-04-03
Just skip the install in test mode

### LINKED ISSUE #290: Don't nuke drives in test mode
Because yeah. We should be able to test the installer screens without nuking drives.

cassidyjames:
Minimally-invasive way to fix #290. Doesn't yet allow us to actually test the progress view (could probably just add a fake timout or something for that), but avoids nuking drives.

### DIFF
--- a/src/Views/ProgressView.vala
+++ b/src/Views/ProgressView.vala
@@ -317,7 +317,12 @@ public class ProgressView : AbstractInstallerView {
         }
 
         new Thread<void*> (null, () => {
-            installer.install ((owned) disks, config);
+            if (Installer.App.test_mode) {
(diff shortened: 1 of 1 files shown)


src/Views/ProgressView.vala
```diff
@@ -317,7 +317,12 @@ public class ProgressView : AbstractInstallerView {
         }
 
         new Thread<void*> (null, () => {
-            installer.install ((owned) disks, config);
+            if (Installer.App.test_mode) {
+                on_success ();
```
jackpot51: Thinking more, you may need to put this inside an `Idle.add` due to it being in a thread.

PR #299, opened 2018-04-11
Update Disks logic

cassidyjames:
This is an upstreaming of changes from @mmstick and the pop fork, which gets the installer building again and sets up a lot of the logic needed for the partitioning screen.

### DIFF
--- a/src/Objects/Mount.vala
+++ b/src/Objects/Mount.vala
@@ -1,0 +1,71 @@
+// -*- Mode: vala; indent-tabs-mode: nil; tab-width: 4 -*-
+/*-
+ * Copyright (c) 2016-2018 elementary LLC. (https://elementary.io)
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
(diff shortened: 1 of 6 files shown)


src/Objects/Mount.vala
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    public bool is_valid_root_mount () {
+        return filesystem != Distinst.FileSystemType.FAT16
+            && filesystem != Distinst.FileSystemType.FAT32
+            && filesystem != Distinst.FileSystemType.NTFS;
+    }
+
+    public bool is_lvm () {
+        return (this.flags & LVM) != 0;
```
tintou: Vala has a syntax for checking flags:
`return LVM in flags;`

src/Objects/Mount.vala
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    public bool should_format () {
+        return (this.flags & FORMAT) != 0;
```
tintou: `return FORMAT in flags`

src/Views/DiskView.vala
```diff
@@ -144,7 +139,7 @@ public class Installer.DiskView : AbstractInstallerView {
                 label = model;
             }
 
-            string path = (owned) path_builder.str;
+            string path = Utils.string_from_utf8(disk.get_device_path ());
```
tintou: spacing here

src/Views/ProgressView.vala
```diff
@@ -297,19 +327,102 @@ public class ProgressView : AbstractInstallerView {
...
+
+                if (m.mount_point == "/boot/efi") {
+                    if (m.is_valid_boot_mount ()) {
+                        if (m.should_format ()) {
+                            partition.format_with (m.filesystem);
+                        }
+
+                        partition.set_mount (m.mount_point);
+                        Distinst.PartitionFlag[] flags = { Distinst.PartitionFlag.ESP };
+                        partition.set_flags (flags);
```
tintou: Use `partition.set_flags ({ Distinst.PartitionFlag.ESP });` as done just after

cassidyjames:
@tintou thanks for the quick review! I've addressed your comments.

jackpot51:
:tada:

PR #301, opened 2018-04-11
Make APT non interactive

tintou:


codygarver:
This was originally in the docker container but somehow got reverted when I was re-arranging branches, I've queued a build and we'll see if that resolves this without this change.

PR #302, opened 2018-04-12
Encrypt after disks

### LINKED ISSUE #295: Move encrpytion after disks
After a day-long discussion across both the Pop!_Chat and elementary Slack, we've decided to go with this compromise for the Disks and Encryption views:

1. Move Encryption to after disks. This lets us choose a disk to blow away **or** custom partitioning.
2. If the user chooses custom partitioning, we let them do any encryption within that view (since there are some combinations that might not make sense for encryption, or some combinations with more than one encrypted drive and password). We skip the normal Encrypt view in this case.
3. Keep the destructive actions on the disk (and custom) view. This is still considered the point-of-no-return for the installer, even if we don't start writing to disk right now.
4. Remove the back/cancel buttons from the Encrypt view. We don't let them go back, because we consider the installation started at this point.

This way we keep a few considerations from the previous version:

1. Clear association between the non-reversible destructive action and the drive or partitions being "destroyed." We leave this decision and destructive action on the disk view.
2. Consistent placement of the non-reversible destructive action, whether or not the user is encrypting. We always show it on the last drive-related screen, whether that's the blow-away-a-disk screen, the custom screen, or—in the future—a smarter dual boot screen.
3. We still require explicit user action _outside of the button area_ to continue; the destructive button is disabled until you explicitly choose a drive to blow away (or partitions to reuse), so you can never just click through and start the installation without an extra precaution.

I think we're in agreement here and want to try this out. It does mean a teeny tiny fib (the installation is not actually starting until after the encrypt view), but I think the trade-offs are worth it.

cassidyjames:
Main question here is the buttons in the Encrypt view. Right now the back button is kind of awkward. See https://youtu.be/hIH5Ng6w-qE

Fixes #295; see that issue for more context.

### DIFF
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -88,7 +88,7 @@ public class Installer.MainWindow : Gtk.Dialog {
         stack.add (try_install_view);
         stack.visible_child = try_install_view;
 
-        try_install_view.next_step.connect (() => load_encryptview ());
+        try_install_view.next_step.connect (() => load_encrypt_view ());
     }
 
     private void set_check_view_visible (bool show) {
@@ -101,7 +101,7 @@ public class Installer.MainWindow : Gtk.Dialog {
         }
     }
 
-    private void load_checkview () {
+    private void load_check_view () {
         if (check_view != null) {
             check_view.destroy ();
         }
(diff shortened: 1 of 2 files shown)


cassidyjames:
@danrabbit I know you did the initial EncryptView, so you might have thoughts.

cassidyjames:
I think this is blocked by #302

PR #308, opened 2018-04-14
Upstream PartitioningView

cassidyjames:
Brings the PartitioningView and related widgets by @mmstick upstream :tada:

### DIFF
--- a/src/Objects/Mount.vala
+++ b/src/Objects/Mount.vala
@@ -1,6 +1,6 @@
 // -*- Mode: vala; indent-tabs-mode: nil; tab-width: 4 -*-
 /*-
- * Copyright (c) 2016-2018 elementary LLC. (https://elementary.io)
+ * Copyright (c) 2018 elementary LLC. (https://elementary.io)
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
@@ -24,6 +24,7 @@ public class Installer.Mount {
     public string mount_point;
     public Distinst.FileSystemType filesystem;
     public Flags flags;
+    public PartitionMenu menu;
 
     [Flags]
     public enum Flags {
@@ -33,11 +34,13 @@ public class Installer.Mount {
     }
 
     public Mount (string partition, string parent_disk, string mount,
-                  Flags flags, Distinst.FileSystemType fs) {
+                  Flags flags, Distinst.FileSystemType fs,
+                  PartitionMenu menu) {
         filesystem = fs;
         mount_point = mount;
         partition_path = partition;
         this.flags = flags;
+        this.menu = menu;
         this.parent_disk = parent_disk;
     }
 
--- a/src/Views/PartitioningView.vala
+++ b/src/Views/PartitioningView.vala
@@ -66,7 +66,6 @@ public class Installer.PartitioningView : AbstractInstallerView {
         );
 
         var description = new Gtk.Label (full_description);
-        description.halign = Gtk.Align.FILL;
         description.margin_bottom = description.margin_bottom = 24;
         description.max_width_chars = 72;
         description.use_markup = true;
--- a/src/Views/PartitioningView.vala
+++ b/src/Views/PartitioningView.vala
@@ -248,8 +248,7 @@ public class Installer.PartitioningView : AbstractInstallerView  {
                 stderr.printf ("decrypt: unable to locate LUKS partition at %s\n", device);
                 break;
             default:
-                // FIXME: Use `critical` as this sounds like a big deal
-                stderr.printf ("decrypt: unhandled error value: %d\n", result);
+                critical ("decrypt: unhandled error value: %d\n", result);
                 break;
         }
     }
--- a/src/Views/PartitioningView.vala
+++ b/src/Views/PartitioningView.vala
@@ -148,7 +148,7 @@ public class Installer.PartitioningView : AbstractInstallerView  {
             var process = new GLib.Subprocess.newv ({"gparted"}, GLib.SubprocessFlags.NONE);
             process.wait ();
         } catch (GLib.Error error) {
-            stderr.printf ("critical error occurred when executing gparted\n");
+            critical ("could not execute gparted");
         }
 
         reset_view ();
@@ -186,12 +186,10 @@ public class Installer.PartitioningView : AbstractInstallerView  {
         const uint8 ROOT = 1;
         const uint8 BOOT = 2;
 
-        // FIXME: store everything in a string and use `debug`, which allows a
-        // fine-grained level of debug outputs
-        stderr.printf ("DEBUG: Current Layout:\n");
+        debug ("Current Layout:");
         foreach (Mount m in mounts) {
-            stderr.printf (
-                "  %s : %s : %s: format? %s\n",
+            debug (
+                "  %s : %s : %s: format? %s",
                 m.partition_path,
                 m.mount_point,
                 Distinst.strfilesys (m.filesystem),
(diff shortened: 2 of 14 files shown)


cassidyjames:
I think this is blocked by #302

cassidyjames:
The one issue I know of and haven't dug into is the action area on the PartitioningView. Something is screwy there and causing it to move compared to the other views.

src/Views/PartitioningView.vala
```diff
@@ -0,0 +1,302 @@
...
+        foreach (unowned Distinst.LvmDevice disk in disks.list_logical ()) {
+            add_logical_disk (disk);
+        }
+
+        disk_list.show_all ();
+    }
+
+    private void open_gparted () {
+        try {
+            var process = new GLib.Subprocess.newv ({"gparted"}, GLib.SubprocessFlags.NONE);
```
tintou: Giving that gparted is providing a .desktop file, I really prefer the use of https://valadoc.org/gio-2.0/GLib.AppInfo.html

src/Views/PartitioningView.vala
```diff
@@ -0,0 +1,302 @@
...
+        label_sizer.add_widget (disk_bar.label);
+        disk_list.pack_start (disk_bar);
+    }
+
+    private void validate_status () {
+        uint8 flags = 0;
+        const uint8 ROOT = 1;
+        const uint8 BOOT = 2;
+
+        stderr.printf ("DEBUG: Current Layout:\n");
```
tintou: please store everything in a string and use the `debug` function which allows a fine-grained level of debug outputs

src/Views/PartitioningView.vala
```diff
@@ -0,0 +1,302 @@
...
+                stderr.printf ("decrypt: unable to decrypt partition (possibly invalid password)\n");
+                break;
+            case 5:
+                stderr.printf ("decrypt: the decrypted partition does not have a LVM volume on it\n");
+                break;
+            case 6:
+                stderr.printf ("decrypt: unable to locate LUKS partition at %s\n", device);
+                break;
+            default:
+                stderr.printf ("decrypt: unhandled error value: %d\n", result);
```
tintou: also here, please use `critical` as it sounds like a big deal

src/Widgets/DiskBar.vala
```diff
@@ -0,0 +1,254 @@
...
+        var context = fill_round.get_style_context ();
+        context.add_class ("legend");
+        context.add_class (fs);
+
+        var format_size = GLib.format_size (size);
+        var info = new Gtk.Label (
+            (vg == null)
+                ? "%s (%s)".printf (format_size, fs)
+                : "%s (%s: <b>%s</b>)".printf (format_size, fs, vg)
+        );
```
tintou: please translate these strings

src/Widgets/PartitionMenu.vala
```diff
@@ -0,0 +1,355 @@
...
+        format_label = new Gtk.Label ("Format:");
+        format_label.halign = Gtk.Align.END;
+        format_label.xalign = 1;
+        label_size_group.add_widget (format_label);
+
+        format_partition = new Gtk.Switch ();
+        format_partition.halign = Gtk.Align.START;
+        format_partition.hexpand = false;
+
+        var use_as_label = new Gtk.Label ("Use as:");
```
tintou: please translate this string

src/Widgets/PartitionMenu.vala
```diff
@@ -0,0 +1,355 @@
...
+        use_partition_label.xalign = 1;
+        label_size_group.add_widget (use_partition_label);
+
+        use_partition = new Gtk.Switch ();
+        use_partition.halign = Gtk.Align.START;
+        use_partition.hexpand = false;
+
+        var separator = new Gtk.Separator (Gtk.Orientation.HORIZONTAL);
+
+        format_label = new Gtk.Label ("Format:");
```
tintou: please translate this string

src/Widgets/PartitionMenu.vala
```diff
@@ -0,0 +1,355 @@
...
+        use_as.active = 0;
+
+        custom_label = new Gtk.Label ("Custom:");
+        custom_label.halign = Gtk.Align.END;
+        custom_label.xalign = 1;
+        label_size_group.add_widget (custom_label);
+
+        custom = new Gtk.Entry ();
+
+        type_label = new Gtk.Label ("Filesystem:");
```
tintou: please translate this string

src/Widgets/PartitionMenu.vala
```diff
@@ -0,0 +1,355 @@
...
+                set_format_sensitivity ();
+                disable_signals = false;
+                use_as.set_active (
+                    (fs == Distinst.FileSystemType.FAT16 || fs == Distinst.FileSystemType.FAT32)
+                        ? mount_set (boot_partition) ? 4 : 2
+                        : fs == Distinst.FileSystemType.SWAP
+                            ? 3
+                            : mount_set ("/")
+                                ? mount_set ("/home" ) ? 4 : 1
+                                : 0
```
tintou: I'm not a fan of this kind of syntax at all, it's the role of the compiler to optimize it, please use temporary variables

src/Widgets/PartitionMenu.vala
```diff
@@ -0,0 +1,355 @@
...
+            custom_label.visible = visible;
+
+            if (active == 2) {
+                if (Distinst.bootloader_detect () == Distinst.PartitionTable.GPT) {
+                    type.active = 2;
+                } else {
+                    type.active = 0;
+                }
+                type_label.visible = true;
+                type.visible = true;
```
tintou: you might want to use `bind_property` between `type` and `type_label` at creation time as they seem to share their visible property and I think that there are others here. it could just make it a little bit more lightweight

src/Widgets/PartitionMenu.vala
```diff
@@ -0,0 +1,355 @@
...
+            this
+        ));
+
+        var mount_icon = new Gtk.Image.from_icon_name (
+            "process-completed-symbolic",
+            Gtk.IconSize.SMALL_TOOLBAR
+        );
+        mount_icon.halign = Gtk.Align.END;
+        mount_icon.valign = Gtk.Align.END;
+        mount_icon.margin = 2;
```
tintou: we always use multiple of `3` for margins

src/Views/PartitioningView.vala
cassidyjames: @tintou does that let us still wait and refresh the partitions when it exits?

cassidyjames:
@mmstick when you get a chance, can you take a look at the remaining FIXMEs in this PR?

src/Objects/Mount.vala
```diff
@@ -1,6 +1,6 @@
 // -*- Mode: vala; indent-tabs-mode: nil; tab-width: 4 -*-
 /*-
- * Copyright (c) 2016-2018 elementary LLC. (https://elementary.io)
```
danrabbit: but why

cassidyjames: Because this was written in 2018, and incorrectly had 2016 in there before

cassidyjames:
@tintou I think I addressed all of your comments. Thanks again for the review! I know it was a doozy.

src/Widgets/DiskBar.vala
```diff
@@ -0,0 +1,256 @@
...
+        var context = fill_round.get_style_context ();
+        context.add_class ("legend");
+        context.add_class (fs);
+
+        var format_size = GLib.format_size (size);
+
+        var info = new Gtk.Label (
+            (vg == null)
+                ? _("%s (%s)".printf (format_size, fs))
+                : _("%s (%s: <b>%s</b>)".printf (format_size, fs, vg))
```
tintou: this won't work, it needs to be
_("%s (%s)").printf (format_size, fs)
_("%s (%s: <b>%s</b>)").printf (format_size, fs, vg)
otherwise it will try to translate the replaced string which won't be included in the translation catalog

cassidyjames: d'oh, thanks

PR #311, opened 2018-04-20
Depend on gparted

danrabbit:
Make sure we depend on gparted since this is used by the custom partition view


cassidyjames:
Looks sane to me, but I'm also not usually one to touch packaging. :smiley:

PR #314, opened 2018-04-23
Rework decrypt menu UI

cassidyjames:
- Changes to a more dialog-style layout
  - Adds an icon + overlay to show it's a locked disk
  - Adds some explanatory copy
- Moves the placeholder text out to labels
- Changes "Physical Volume" to "Device name"
- Sets a default physical volume name to `data`
- Adds a suggested action class to the button
- Sets the button to insensitive by default

| `master` | `decrypt-popover` |
|------------|-------------------------|
| screenshot from 2018-04-23 10 10 47 | screenshot from 2018-04-23 11 04 50 |
| screenshot from 2018-04-23 10 10 59 | screenshot from 2018-04-23 11 05 00 |

### DIFF
--- a/src/Widgets/DecryptMenu.vala
+++ b/src/Widgets/DecryptMenu.vala
@@ -37,6 +37,37 @@ public class Installer.DecryptMenu: Gtk.Popover {
     }
 
     private void create_decrypt_view (string device_path, DecryptFn decrypt) {
+        var image = new Gtk.Image.from_icon_name ("drive-harddisk", Gtk.IconSize.DIALOG);
+        image.valign = Gtk.Align.START;
+
+        var overlay_image = new Gtk.Image.from_icon_name ("locked", Gtk.IconSize.DND);
+        overlay_image.halign = Gtk.Align.END;
+        overlay_image.valign = Gtk.Align.END;
+
+        var overlay = new Gtk.Overlay ();
+        overlay.halign = Gtk.Align.CENTER;
+        overlay.valign = Gtk.Align.END;
+        overlay.width_request = 60;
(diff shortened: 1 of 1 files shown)


cassidyjames:
I kept this a popover for now so that it's still closely tied to the partition the user is working with; I feel like pulling it into a dialog would require more copy to associate the dialog to the partition.

cassidyjames:
I also debated using the large password icon and badging it with a disk, but on second thought I didn't like that: we use that pattern for auth dialogs where the user must enter their password, and this is a different context with a different password. They're unlocking the disk using its password, not a user password.

If that distinction is unimportant, we could switch it to the more auth-like icon. But I thought I'd share my thought process there.

danrabbit:
Documenting discussion in slack:

- [ ] Title Case primary label
- [ ] `dialog-password` instead of lock icon to indicate that we're entering a password to _unlock_ and not locking

src/Widgets/DecryptMenu.vala
```diff
@@ -21,35 +21,72 @@
...
+
+        var overlay = new Gtk.Overlay ();
+        overlay.halign = Gtk.Align.CENTER;
+        overlay.valign = Gtk.Align.END;
+        overlay.width_request = 60;
+        overlay.add (image);
+        overlay.add_overlay (overlay_image);
+
+        var primary_label = new Gtk.Label (_("Decrypt This Partition"));
+        primary_label.get_style_context ().add_class ("primary");
```
danrabbit: Oops, this should be `Granite.STYLE_CLASS_PRIMARY_LABEL`

PR #321, opened 2018-04-23
Simplify decryption copy

### LINKED ISSUE #319: Don't use the word "LUKS"
From Google+:

screenshot from 2018-04-23 14 06 32 2x

We probably shouldn't expect users to know what LUKS is

cassidyjames:
image

Fixes #319


danrabbit:
Actually, looking at this I wonder if "Device name" was better since you know... name and password. It might be kind of confusing thinking this is your username

cassidyjames:
@danrabbit okay, then how about "Enter the partition's encryption password and set a device name for the decrypted partition." as well?

cassidyjames:
@danrabbit addressed

PR #326, opened 2018-05-01
Check if we're in a VM

### LINKED ISSUE #323: Warn about VMs
Stolen from deepin. They throw a warning screen if you’re installing on a VM about performance etc. Probably a good idea 

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/57647282-warn-about-vms?utm_campaign=plugin&utm_content=tracker%2F46189108&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F46189108&utm_medium=issues&utm_source=github).
</bountysource-plugin>

cassidyjames:
Fixes #323

- Update copyright date while we're here
- Add a check to see if we're in a VM based on `cpuinfo` flags
- Change `enough_power` to `minimum_specs` to reduce confusion
- Move `encrypt_view` to its logical place in MainWindow.vala
- Revert a lot of the check view magic as it appears to be broken with multiple conditions (i.e. on battery and in a VM)

Screenshot from my VM:

image

Ready for review.


cassidyjames:
>I think there's a better way of checking that through CPU kernel flags file… There's a `hypervisor` flag in `/proc/cpuinfo` when we're running on a VM
–@donadigo

cassidyjames:
Confirmed to work in VirtualBox here, should probably double-check other VM software for sanity.

cassidyjames:
Spinning up a VMware machine for testing, though with the distinst changes that might be difficult.

cassidyjames:
Confirmed to work in VMware. Now the only problem I see is that this only loads when the user chooses a clean install, but not a custom partition.

cassidyjames:
Fixed the previous comment, but it feels a little dirty. Help/input welcome! :grin:

jeremypw:
@cassidyjames Converting to draft as labelled "In Progress", has conflicts, and no recent progress.

cassidyjames:
Closing in favor of #625

PR #329, opened 2018-05-15
Install type

### LINKED ISSUE #328: Simple InstallType view
Something that would move us in the direction of #320 without much more work would be to start building the InstallType view where we can set up possible options, like the "Refresh," "Upgrade," etc. magic partitioning types. 

For now, it could simply support the "Clean Install" and "Custom…" types since those are already supported. Then when we do Home folder retention, it gives us a natural place to put it.

cassidyjames:
Fixes #328 by changing the TryInstallView to be an installation type view. Input welcome on copy, but it's a start and does not lose any functionality. It gives us a place to hook up things like Preserve Home Folder, Reinstall, Refresh, smart Dual Boot, etc.

screenshot

Largely inspired by the DiskView and the [old spec](https://docs.google.com/document/d/1Sw07eNjORV1rBEGhlWJmD39BgNBVNbJEJhLzrc_6T0w/edit#)'s "Installation Options" view:

old mockup


src/Widgets/InstallTypeGrid.vala
```diff
@@ -0,0 +1,68 @@
...
+        get_style_context ().add_class (Gtk.STYLE_CLASS_FLAT);
+
+        var type_image = new Gtk.Image.from_icon_name (icon_name, Gtk.IconSize.DIALOG);
+        type_image.use_fallback = true;
+
+        var title_label = new Gtk.Label (type_title);
+        title_label.halign = Gtk.Align.START;
+        title_label.hexpand = true;
+        title_label.valign = Gtk.Align.END;
+        title_label.get_style_context ().add_class ("primary");
```
danrabbit: should be `Granite.STYLE_CLASS_PRIMARY_LABEL` probably

src/Views/TryInstallView.vala
```diff
@@ -16,99 +16,147 @@
...
-        grid.row_spacing = 32;
-
-        grid.attach (choice_image, 0, 1, 1, 1);
-        grid.attach (choice_label, 1, 1, 1, 1);
-        grid.attach (implications_image, 0, 2, 1, 1);
-        grid.attach (implications_label, 1, 2, 1, 1);
-        grid.attach (return_image, 0, 3, 1, 1);
-        grid.attach (return_label, 1, 3, 1, 1);
+    private Gtk.Button next_button;
+    private Gtk.Grid type_grid;
```
danrabbit: This doesn't need to be declared here since it's only used inside the construct block. You can just do `var type_grid`

PR #332, opened 2018-05-24
Distinst string → utf8 changes

cassidyjames:
Fixes broken builds due to changes in distinst

### DIFF
--- a/src/Views/DiskView.vala
+++ b/src/Views/DiskView.vala
@@ -132,9 +132,9 @@ public class Installer.DiskView : AbstractInstallerView {
             }
 
             string label;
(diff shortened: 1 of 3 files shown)


mmstick:
elementary may also want to import the improvements done to to locales (lang + country) and keyboard layouts in the pop-os fork. It removes a lot of code from the frontend since it's now handled in distinst, fixes issues with various locale configurations that cause installs with certain combinations to fail, and provides many more langs/countries to select from. Every language should have at least one country to choose from (or a None option for languages like Esperanto).

jackpot51:
I wanted to work on the sync PR's but have been too busy

PR #333, opened 2018-05-25
Decrypt Before Try/Install

cassidyjames:
WIP. The goal is to add an infobar on the Try/Install view with an unlock button if encrypted partitions are detected. Then to use a new decryption dialog (which should ideally replace the existing popover from the custom partitioning view) to unlock any encrypted partitions.

image

img_20180525_125231

## TODO:

- [ ] Styling for the infobar in a dialog? This is a little weird, but it seems like the right pattern to use.
- [x] Finish the DecryptDialog
- [x] Fix that shutdown button…


src/Views/TryInstallView.vala
```diff
@@ -55,14 +55,32 @@ public class Installer.TryInstallView : AbstractInstallerView {
...
+        title_grid.attach (type_label,      0, 1);
+        title_grid.attach (type_desc_label, 0, 2);
+
+        // TODO: Once we support more options, give an example here
+        // ("More options, such as…") if there's space…
+        var decrypt_description = new Gtk.Label (_("More options may be available after unlocking encrypted drives"));
+
+        var decrypt_button = new Gtk.Button.with_label (_("Unlock Encrypted Drives…"));
+
+        // TODO: Only show decrypt infobar when we detect encrypted drives
```
mmstick: In https://github.com/pop-os/installer/pull/126, this will soon be possible through `InstallOptions.get_default ().contains_luks ()`. Then for getting a list of devices to decrypt, I'll add a `borrow_disks ()` method which you could use to check partitions with a file system of `LUKS`, and for attempting the decryption.

cassidyjames:
@mmstick I think I'm ready for logic here. I could see some copy being tweaked but I'm pretty happy with the UI itself.

mmstick:
I'm going to check this out and see if it I can get it working.

mmstick:
I'm having some early success. The dialog is decrypting partitions and updating possible installation options. It's able to find the OS on a partition within an encrypted device and get that partition's UUID:

```
found Pop!_OS on c9d68076-b5d0-4d61-aa64-285c8f88d28a
```

Now I just need to add a new button for enabling refresh installations on that retains `/home`.

jeremypw:
@cassidyjames Converting to draft as conflicts with master and no recent activity.

PR #334, opened 2018-06-15
Use GTK to inhibit sleep

mdh34:
Upon testing Gtk.Application.inhibit works great


Philip-Scott:
@cassidyjames Maybe it works now, but it may have not existed /working back in the Loki days

cassidyjames:
Neither are actually working here in GNOME Shell, and I believe our Juno builds don't have automatic locking/suspend actually working. So I'm not sure what to do or how to properly test. :stuck_out_tongue:

mdh34:
closing this due to inactivity

cassidyjames:
@Philip-Scott I know you looked at inhibiting again recently in Spice-Up. How does this compare?

mmstick:
We found out that inhibition doesn't work when the session is driven by root, as is the case for the installer's session. All inhibitors are ignored.

Philip-Scott:
@mmstick Not sure if this is the same case when it's in root, but at least in my tests i found out that i needed to inhibit at two separate places to prevent a suspend and the screen from turning off. And in one place i had to send it specific flags

This is where inhibiting occurs within Spice-Up https://github.com/Philip-Scott/Spice-up/blob/master/src/GraniteStaging/Inhibitor.vala#L58

PR #338, opened 2018-06-15
Help Dialog

mmstick:
Split from #335


cassidyjames:
Original author of this (terrible) widget here! I'm not super happy with this for a few reasons:

- [ ] It seems difficult to maintain. I'm not sure if this text should be loaded from elsewhere, or what
- [ ] It's not currently translatable
- [ ] Right now it allows infinite instances of the help dialog. We can't make it blocking because the user might be referring to it while configuring their drive, so we need to only allow one instance of it.

I was also playing with the idea of splitting it up into smaller sections with a sidebar, but that might be something for another PR.

Issue #253, opened 2018-03-09
Different keyboard layouts can not be tested on KeyboardLayoutView page


brs17:
When selecting a keyboard that isn't the standard US QWERTY keyboard, typing in the `Type to test your layout` input still outputs in QWERTY.

I have tried both Spanish and English dvorak keyboard layouts.

joshuajaharwood:
The same happens when testing UK QWERTY on 6.0 Next, build 8/9/20; I'm stuck on the US layout. This is fixed once we boot from the install drive into user setup.

danrabbit:
Hm, this should have been fixed in https://github.com/elementary/installer/pull/443. It looks like the latest build of installer finished less than 20 hours ago, so this fix might be in the September 9th daily OS image

carniz:
Still seeing this with the 20200914 build when trying to use a Swedish keyboard layout.

carniz:
...and still present in elementaryos-6.0-daily.20201128.iso, so I guess [#443](https://github.com/elementary/installer/pull/443) didn't do the trick :/

carniz:
...and 2 months later it's still there (tested elementaryos-6.0-daily.20210118.iso)

carniz:
To be clarified: the issue is still there in the pre-restart phase of the installation (i.e. prior to setting up a user):

![VirtualBox_elementary6install_31_01_2021_11_31_44](https://user-images.githubusercontent.com/1648530/106381643-81f02d00-63ba-11eb-8ab1-e521e6f7618b.png)

After the restart the keyboard layout works as it should:

![VirtualBox_elementary6install_31_01_2021_11_45_20](https://user-images.githubusercontent.com/1648530/106381663-a51adc80-63ba-11eb-997f-8ddc31e896f3.png)

PR #340, opened 2018-06-18
Rewrite Locale Support

mmstick:
- Use distinst to get country and language codes
- Many more languages and countries are now supported
- Failures due to invalid locale configurations are no longer possible
- No more hard-coded values or hand-written XML parsing
- Keyboard layouts may now be previewed

## Before

English, which seems to ignore a significant number of countries that speak English.

screenshot from 2018-06-21 16-43-30

## After

All the countries with English language support. The Irish no longer have to select their bitter rival, the UK, as their locale.

screenshot from 2018-06-21 16-44-15

## Countries for languages that previously had none

screenshot from 2018-06-21 16-47-56

Requires #342 & #341 to pass Travis checks
Closes https://github.com/elementary/installer/issues/253


tintou:
wait, what ?! what does the Languages and Layout do in distinst ?

mmstick:
@tintou  All of the source code relating to keyboard layouts, ISO codes, etc. are in the [locale](https://github.com/pop-os/distinst/tree/master_bionic/src/locale) module of distinst.

  - There's an `i18n` submodule which creates a static map of supported locale variants (country and language codes, + codesets) from `/usr/share/i18n/SUPPORTED`.
  - The `main_countries` submodule creates a static map of preferred countries for each language code, parsing the `/usr/share/language-tools/main-countries` file.
  - [Serde](https://serde.rs/) is used to automatically deserialize `/usr/share/X11/xkb/rules/base.xml` into native Rust data structures within the `keyboard_layout` submodule, for obtaining keyboard layouts.
  - The `iso_codes` submodule is responsible for converting language & country codes into their names, as welll as providing the option to get translated variants of their names.
    - The [isolang](https://crates.io/crates/isolang) crate is responsible for handling ISO 639 language codes.
    - The [iso3166-1](https://crates.io/crates/iso3166-1) crate is responsible for handling ISO 3166-1 country codes.
    - The [gettext-rs](https://crates.io/crates/gettext-rs) crate provides translation support for both of these codes.

mmstick:
Updated description with some screenshots.

mmstick:
@tintou The only languages & country combinations that are listed are those that are within `/etc/i18n/SUPPORTED`. Before, the UI was capable of selecting options which are not supported, hence causing `localectl` to fail. Some language choices, such as Portugese, would cause the installer to fail; and others, such as Spanish, could not use a Spanish keyboard layout.

> United Kingdom of Great Britain and Northern Ireland

[That's officially what the ISO 3166-1 spec defines as the English 'short name'](https://en.wikipedia.org/wiki/ISO_3166-1#Officially_assigned_code_elements), but we can override them with even shorter names.

cassidyjames:
@tintou what are your current thoughts on this? My main concern is not matching the language and region settings in either GNOME or elementary OS. However, I also don't know enough about how it works in master (or what an ideal solution would be) to judge this too much.

tintou:
I'm still unsure about this, while giving a list of supported locales makes perfectly sense, I don't think that it's distinst duty to handle anything for the current session.
Also, the "main countries" are unfortunately hardcoded because /usr/share/language-tools/main-countries is an Ubuntu-only thing…
In an ideal world, there would be a separate library to handle the listing of locale codes/names/translated names

mmstick:
@tintou distinst's duty is to handle everything that an installer would need to handle. The UI should not be doing anything outside of calling to distinst to perform each task, and that includes managing language locales and keyboard settings for the current and future session. A theoretical CLI, Qt, or OrbTK front end should not need to re-implement the same installer details.

It's additionally vital for distinst to handle these details, so as to avoid invalid configurations when setting up the locales and keyboard settings within the chroot of the new install. Before this, there were a lot of issue reports about installer errors due to invalid language and keyboard configurations, and language / keyboard settings not being applied. Since the pop installer switched to this code, such issues are no more, so I believe it is the correct solution to the problem.

> In an ideal world, there would be a separate library to handle the listing of locale codes/names/translated names

Distinst is already invoking separate libraries for listing locale codes / names / translations, as mentioned at https://github.com/elementary/installer/pull/340#issuecomment-398146876. They're based on their ISO specifications, so there's no room for error. They are statically linked into the distinst library at compile-time. The installer itself could statically link distinst too, if it wanted to.

tintou:
> The UI should not be doing anything outside of calling to distinst to perform each task, and that includes managing language locales and keyboard settings for the current and future session. A theoretical CLI, Qt, or OrbTK front end should not need to re-implement the same installer details.

That's exactly why we need a separate library for this (and why it should remain client-side for now), distinst duty is to give a list of supported locale codes. My major concern here is that we might want distinst to run in another process as the GUI shouldn't run as root on Wayland.

jlnr:
@mmstick The unicode-cldr-core package is only 10 MB. Is the rust packaging the issue (I couldn't find a library for the CLDR country names)?
Should I move the ISO-3316 discussion to a new issue on another project?

toto6038:
> 
> 
> Sorry to jump into this with politics:
> 
> > That's officially what the ISO 3166-1 spec defines as the English 'short name', but we can override them with even shorter names.
> 
> ISO 3166-1 country names are a mess. Many are cumbersome, and some are politically loaded. It sucks, but ISO (via the UN) is not politically neutral.
> 
> I would strongly suggest using [Unicode CLDR data](http://cldr.unicode.org/translation/country-names) for anything that will be displayed to the end user.
> 
> [This Drupal issue](https://www.drupal.org/project/drupal/issues/1938892) on migrating from ISO 3166-1 to CLDR goes into detail on some of the reasons, besides what is listed on the CLDR page above.

You're right. The usage of troublesome ISO 3166-1 list results in the situation that I have to choose "Taiwan, Province of China", a name that doesn't comply to truth and most civilians on the island don't recognize to, during installation process. I would rather just choose English instead. :(

PR #345, opened 2018-06-18
Perform additional mount validation checks

mmstick:
- Ensure that the EFI partition is large enough
- Check that key mounts have a valid file system
- Return errors so that the UI can display the errors

### DIFF
--- a/src/Views/PartitioningView.vala
+++ b/src/Views/PartitioningView.vala
@@ -38,6 +38,8 @@ public class Installer.PartitioningView : AbstractInstallerView {
         Object (cancellable: false);
     }
 
+    const uint64 REQUIRED_EFI_SECTORS = 524288;
+
     construct {
         mounts = new Gee.ArrayList<Installer.Mount> ();
         luks = new Gee.ArrayList<LuksCredentials> ();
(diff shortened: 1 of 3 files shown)


src/Views/PartitioningView.vala
```diff
@@ -190,34 +192,38 @@ public class Installer.PartitioningView : AbstractInstallerView {
...
+            stderr.printf (
+                "  %s : %s : %s : %s: format? %s\n",
+                m.parent_disk,
                 m.partition_path,
                 m.mount_point,
                 Distinst.strfilesys (m.filesystem),
                 m.should_format () ? "true" : "false"
             );
-        }
-        debug ("Current Layout:\n" + layout_debug);
```
tintou: revert to the original debugging output

src/Views/PartitioningView.vala
```diff
@@ -190,34 +192,38 @@ public class Installer.PartitioningView : AbstractInstallerView {
...
-            if (flags == ROOT + BOOT) {
-                next_button.sensitive = true;
-                return;
+            } else if (m.mount_point == "/boot/efi") {
+                flags |= EFI;
             }
         }
 
-        next_button.sensitive = false;
+        next_button.sensitive = ((flags & required) == required);
```
tintou: next_button.sensitive = required in flags

src/Views/PartitioningView.vala
```diff
@@ -190,34 +192,38 @@ public class Installer.PartitioningView : AbstractInstallerView {
     private void validate_status () {
         uint8 flags = 0;
         const uint8 ROOT = 1;
-        const uint8 BOOT = 2;
+        const uint8 EFI = 2;
```
tintou: It's really time to create an:

[flags]
public enum Status {
    ROOT,
    EFI
}

src/Views/PartitioningView.vala
```diff
@@ -258,19 +264,33 @@ public class Installer.PartitioningView : AbstractInstallerView {
         }
     }
 
-    private void set_mount (Mount mount) {
+    private string? set_mount (Mount mount) {
```
tintou: To me all the returned strings look like errors,
better do a `private void set_mount (Mount mount) throws GLib.Error` and then `throw new GLib.IOError.FAILED (_("blah"));`

PR #348, opened 2018-06-18
Improve DiskBar

mmstick:
Diff between Pop!_OS installer & elementary installer.

- Partitions now have an accompanying legend to identify the partition
- Partition bars have a better minimum size, to accomodate icons
- Improved allocation of partition bar widths


src/Widgets/DiskBar.vala
```diff
@@ -174,7 +180,7 @@ public class Installer.DiskBar: Gtk.Grid {
                 // Begin by resizing all partitions over 20px wide.
                 bool excess_modified = false;
                 for (int y = 0; excess > 0 && y < x; y++) {
-                    if (lengths[y] <= 20) continue;
+                    if (lengths[y] <= 28) continue;
```
cassidyjames: As long as we're modifying this, does it make sense to use a CONST to keep it all in sync?

src/Widgets/DiskBar.vala
```diff
@@ -174,7 +180,7 @@ public class Installer.DiskBar: Gtk.Grid {
                 // Begin by resizing all partitions over 20px wide.
```
cassidyjames: Comment references old value

src/Widgets/DiskBar.vala
```diff
@@ -207,6 +213,10 @@ public class Installer.DiskBar: Gtk.Grid {
     }
 
     internal class FillRound : Gtk.Widget {
+        internal FillRound () {
```
cassidyjames: Is this empty method needed?

mmstick: No idea. At the time of writing this code, I simply copied the pre-existing storage bar Vala code to use as a template for implementing the disk bar, which had this empty method.

src/Widgets/DiskBar.vala
mmstick: Sure, I'll do that.

src/Widgets/DiskBar.vala
```diff
@@ -24,91 +24,69 @@ public class Installer.DiskBar: Gtk.Grid {
...
-        label.margin_end = 12;
-        label.valign = Gtk.Align.CENTER;
-        label.add (name_label);
-        label.add (size_label);
+        unused = get_unused ();
+        generate_label ();
+        generate_bar ();
+        generate_legend ();
+
+        this.hexpand = true;
```
cassidyjames: Probably don't need to explicitly namespace these with `this`, it's clear without.

cassidyjames:
It looks like the empty space is not filled with a `.fill-block`, which is how Granite.StorageBar and the stylesheet work. We could look at updating the stylesheet to give the `.trough` a better default style, but considering this is a regression from before I think I'd have to be sold. I know the StorageBar widget and stylesheet were written in a very specific way to get the exact styleability we want, with the empty portion being another block on the same level as the others (affecting rounded corners and shadows in the styling) versus the blocks being visually on-top of the empty space.

cassidyjames:
Here is a better look.

**master:**

screenshot from 2018-08-31 10 30 06

**this branch:**

screenshot from 2018-08-31 10 29 29

jeremypw:
@mmstick Closing as has unaddressed issues and no recent activity. Please feel free to reopen (as draft) if you wish to continue working on it.

PR #349, opened 2018-06-19
Allow specifying a view

cassidyjames:
Adds a `--load` argument, e.g. `io.elementary.installer -t --load=error`. Useful for testing specific views. Input welcome.


src/Application.vala
```diff
@@ -21,10 +21,12 @@
 public class Installer.App : Gtk.Application {
     public const OptionEntry[] INSTALLER_OPTIONS = {
         { "test", 't', 0, OptionArg.NONE, out test_mode, "Non-destructive test mode", null},
+        { "load", 'l', 0, OptionArg.STRING, out load_view, "Force load a view", "VIEW"},
```
danrabbit: I guess this should probably list all the valid options right?

cassidyjames: Yeah that's sane. I wonder if we can dynamically construct the options instead of just maintaining lists in multiple places?

cassidyjames: @danrabbit any further thoughts on this, it's hard to test the error view as-is. :stuck_out_tongue:

cassidyjames:
@danrabbit ah yeah I was always just trying with `--test` as well I can probably just set it, too.

cassidyjames:
@danrabbit I added a check for `load_view != null` everywhere we're using `test_mode`, but that feels really verbose and prone to missing it later. Suggestions welcome.

jeremypw:
@cassidyjames : Converting to draft as has conflicts and/or no recent activity.  Please feel free to close or reopen for review after fixing conflicts as desired.

PR #350, opened 2018-06-21
Improved keyboard navigation support

### LINKED ISSUE #279: Keyboard accessibility
**Current Behavior**
When selecting a language, <kbd>Enter</kbd> will by default select English and take you to select the type of English. Selecting <kbd>Enter</kbd> again does not do anything. To be able to advance to the next page selecting <kbd>Tab</kbd> four times is necessary.

This same incorrect functionality happens on the keyboard layout page as well.

**Expected Behavior**
When the language is selected, <kbd>Enter</kbd> should progress to the next screen.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/56463698-keyboard-accessibility?utm_campaign=plugin&utm_content=tracker%2F46189108&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F46189108&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #280: Keyboard progression after drive encryption
**Current Behavior**
After typing in and confirming a password for drive encryption, <kbd>Enter</kbd> does not do anything. Furthermore pressing <kbd>Tab</kbd> first highlights the `Don't Encrypt` button.

**Expected behavior**
After typing in and confirming a password for drive encryption, <kbd>Enter</kbd> should progress the installer to the drive selection page. Pressing <kbd>Tab</kbd> should highlight the `Select Drive` button (especially since <kbd>Enter</kbd> does not work)

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/56398651-keyboard-progression-after-drive-encryption?utm_campaign=plugin&utm_content=tracker%2F46189108&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F46189108&utm_medium=issues&utm_source=github).
</bountysource-plugin>

mmstick:
- The return key will now select the active option and move to the next view / variant list box.
- The backspace and return keys will logically move backwards to the previous view / unset the variant list box

Closes #279 
Closes #280 

I'm going to keep adding onto this for a bit longer.

Requires #340 to fix an issue with unsetting the variant


tintou:
I'm unsure that we want to give this meaning to `backspace`, we never did and I wonder if it will work properly with entries (untested here)…

tintou:
But otherwise, good move :)

mmstick:
@tintou Yes, backspace it still works with entries.

danrabbit:
I'm -1 on introduce backspace as a navigation shortcut. We've had this discussion several times in other apps and we always decide that it's a bad idea. Even the major browsers are moving away from this shortcut. Instead, the standard backspace shortcut is `alt` + `←`

src/Views/EncryptView.vala
```diff
@@ -185,6 +185,15 @@ public class EncryptView : AbstractInstallerView {
             update_next_button ();
         });
 
+        confirm_entry.key_press_event.connect ((event) => {
```
danrabbit: There is a native Gtk+ way to do this. You set [`activates_default`](https://valadoc.org/gtk+-3.0/Gtk.Entry.activates_default.html) to true and then set `next_button` as [`has_default`](https://valadoc.org/gtk+-3.0/Gtk.Widget.has_default.html)

mmstick:
@danrabbit I'll start working on fixing the issues you pointed out. Shouldn't take long.

danrabbit:
For the back button shortcut we should be using GLib.Action and action accelerators. Manually listening to key press events always gets us in trouble with layouts and such.

For the most part, I don't think it's a good idea to manually grab the enter key. It would probably be better to use the activation signal from the listbox

danrabbit:
Gonna close this since there's not been any movement in quite a few months

mmstick:
@danrabbit Feel free to close all of them. They're vastly outdated compared to our current work, as seen in https://github.com/pop-os/installer/pull/173

PR #365, opened 2018-11-08
Dual boot

cassidyjames:
Mostly just opening this for UI feedback and iteration.

image

Details:
- "Other OS" should be replaced with the other OS's name if we can get it
- The scale is inverted, meaning it starts from the right (in LTR languages). This is handy for a couple of reasons:
  - The value of the scale is just the value of how much space we should make for the new OS
  - The "fill level" of the scale is how large the new OS could be, and automatically constrains the scale to that value so we don't make it too big
  - The filled-in part of the scale is directly correlated with the accent-colored label of the new OS
  - The new OS goes on the right, visually reinforcing that the old OS is already on the disk and the new OS will be effectively appended on
- The scale has a constrain method which keeps it from being sized too small
- The new amount reserved for each partition as well as the resulting free space live updates when moving the scale

To discuss/decide:
- [ ] Manual input with an entry? I'm not convinced this is needed for most people, and if someone wants super nitpicky control, they can always use Custom…
- [ ] Should we clamp to GB, MB, or…? I think displaying it as GB will be good for most people, but could be convinced to show 12.52 GB or whatever. How finicky do we want to get?


mmstick:
May want to ask @WatchMkr about the design

WatchMkr:
@mmstick please post a screenshot of where we are now.

WatchMkr:
@mmstick let's try again with the spinbuttons on their own line like Cassidy's design.

@cassidyjames we tried once but the spin buttons in our theme are... big... and needed a closer look to get it right. What do you think of the size unit being in the spinbutton (GiB) box instead of to the right of the buttons?

cassidyjames:
@WatchMkr I'm not sure that's supported natively by GTK; it's not a pattern I've seen elsewhere at least.

isantop:
@WatchMkr The new version (4.1) should make them quite a bit smaller. They're nearly as small as elementary's now.

mmstick:
I'll have the spin buttons added soon.

WatchMkr:
@isantop is 4.1 backported to 18.04?

mmstick:
screenshot from 2019-01-22 09-47-10

cassidyjames:
I think having the scale above with the OS details and spin buttons below feels a bit more balanced, plus makes it more obvious that the scale is the primary control here—you don't have to get fiddly with exact GBs unless you want to. With them above, they're first in the flow and so might feel like the user has to use them instead of just the scale.

WatchMkr:
With the labels beneath, we felt that there is a disconnect between what your doing with the slider until you read below the slider. Being above certainly doesn't work well with the spinbuttons. @mmstick let's try the OS label above the slider, the spinner and free space below.

mmstick:
Done 
screenshot from 2019-01-22 10-36-00

WatchMkr:
Ha, ewe terrible. Okay, everything below the slider :-).

isantop:
@WatchMkr 4.1 is now in 18.04, yes.

vjr:
@cassidyjames will the installer support "multi boot" not just dual boot? meaning, if there are more than one "other OSes" can elementary be installed safely? something similar to ubiquity's UI?

also, is there an option to *not* install a bootloader - because a user already has another bootloader from another OS installed already configured?

asking because i saw the recent reddit post about daily ISOs with the new installer going to be open to the public at some point in the (near?) future :-)

cassidyjames:
@vjr yes, that's supported by the Custom (Advanced) install option.

jeremypw:
@cassidyjames Converting to draft as has conflicts and/or no recent activity.  Please feel free to close or reopen for review after fixing conflicts as desired.

PR #387, opened 2019-04-30
ErrorView: Redesign Terminal

### LINKED ISSUE #386: Terminal on error view doesn't animate or have proper margins
Probably easier to fix now with #385

danrabbit:
Fixes #386 

Screenshot from 2019-04-30 13 54 46@2x
Screenshot from 2019-04-30 13 54 37@2x

Redesigns this view to be more like what we see in https://github.com/elementary/granite/blob/master/lib/Widgets/MessageDialog.vala

Updates some properties of the terminal widget to match the dialog

### DIFF
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -202,6 +202,7 @@ public class Installer.MainWindow : Gtk.Window {
         stack.visible_child = progress_view;
 
         progress_view.on_success.connect (() => load_success_view ());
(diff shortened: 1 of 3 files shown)


cassidyjames:
@danrabbit oh that's really interesting. Probably makes sense so you can still see the copy and your options. I'll give it a run and look at the code shortly.

Issue #394, opened 2019-08-13
HiDPI not detected


danrabbit:
Not sure what compositor we're running during this session, but it doesn't seem like HiDPI is being detected

cassidyjames:
@danrabbit if I recall correctly from Pop!_OS, this was due to the installer launching too quickly before GSD was fully initialized or something… I believe they just did a dumb timeout to wait, but did we also have this issue in Greeter?

danrabbit:
In the greeter we listen to the scale factor just to maximize the window, but one thing to check is if GSD is running in the session

PR #396, opened 2019-08-14
Start gala-daemon in Installer session

codygarver:


codygarver:
I imagine this is why a lot the Installer session does not function as expected.

BUT I'm not sure ALL of this is necessary.

Needs testing.

codygarver:
Reduced the scope of this PR to only add gala-daemon to the Installer session, this should help hidpi a little bit.

danrabbit:
I'm pretty sure that gala-daemon is just used for native right click menus on window borders. I guess that's not bad to have, but it doesn't solve #394

PR #405, opened 2019-10-02
Fix code style issues

danrabbit:


danrabbit:
@cassidyjames since the DBus name is `UnInhibit` I think we need this to be `un_inhibit` so that it's translated correctly.

Is that right @tintou ?

tintou:
@danrabbit yes

PR #406, opened 2019-10-22
Use casper hook to autostart installer

davidmhewitt:
This changes the `deb-packaging` branch to not overwrite the `lightdm.conf` for the purposes of starting the installer.

We can use `casper`, the tool that autoconfigures users and autologin etc when running in a live session to install the `.desktop` file to the live user's autostart directory. Also helps in the case where someone accidentally installs the installer into their installed system. Their `lightdm.conf` won't get overwritten to boot them into the installer session instead as `casper` isn't there to do the config.

`casper` will automatically detect what display manager is in use and knows how to configure it to automatically log in, so this should be a bit more robust and portable and is consistent with what Pop are doing downstream.

This is not yet tested, but I'm in the process of setting up nightly builds for a distinst based .iso too, so it should be easier to start getting this repo up to scratch soon.

### DIFF
--- a/debian/io.elementary.installer-casper.install
+++ b/debian/io.elementary.installer-casper.install
@@ -1,0 +1,1 @@
+debian/99pop-installer usr/share/initramfs-tools/scripts/casper-bottom/
--- a/debian/io.elementary.installer-casper.install
+++ b/debian/io.elementary.installer-casper.install
@@ -1,1 +1,1 @@
-debian/99pop-installer usr/share/initramfs-tools/scripts/casper-bottom/
+debian/99elementary-installer usr/share/initramfs-tools/scripts/casper-bottom/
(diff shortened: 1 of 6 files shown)


debian/io.elementary.installer-casper.install
```diff
@@ -0,0 +1 @@
+debian/99pop-installer usr/share/initramfs-tools/scripts/casper-bottom/
```
danrabbit: Do we still want this to be named `pop-installer` or was this just a copy/paste thing?

cassidyjames: Looks like copy-paste since the file above is `debian/99elementary-installer`

```suggestion
debian/99elementary-installer usr/share/initramfs-tools/scripts/casper-bottom/
```

davidmhewitt:
After some more investigation, this method isn't going to work for us. The lightdm config that casper injects does not allow guest login, so our switch to demo mode that switches to the guest session won't work.

Pop! have worked around this by running the installer in the actual session and just swapping the switch to guest session code with code that just exits the installer.

PR #410, opened 2020-01-26
AbstractInstallerView: make constructor protected

davidmhewitt:
In our new version of vala, abstract classes can't have public constructors anymore.


tintou:
@davidmhewitt We probably don't even need this constructor altogether

davidmhewitt:
@tintou probably, but I was trying not to change too much in case we have to rebase loads of this on top of the Pop_OS version.

PR #416, opened 2020-02-25
Replace system monitor with tach-slow

### LINKED ISSUE #413: Symbolic performance icon
It looks like we’re missing a symbolic icon representing performance for the encryption view

danrabbit:
Fixes #413 

Screenshot from 2020-02-24 19 58 16@2x


lainsce:
Will this icon eventually be backported for apps that may find it useful?

Issue #442, opened 2020-08-25
Demo Mode should begin in a selected language


ryonakano:
**Describe the bug**
Demo Mode always uses English as the display language regardless of a language you selected in TryInstallView.

**To Reproduce**
1. Boot from ISO
2. Select any language other than English
3. Select "Try Demo Mode" in TryInstallView
4. Demo Mode begins and you'll see the UI is shown in English

**Expected behavior**
Demo Mode should use a language you selected in TryInstallView as the display language, like the old installer Ubiquity.

**Desktop (please complete the following information):**
 - OS: elementary OS 
 - Version: Odin Daily (the ISO as of August 22)

ryonakano:
Confirmed this issue was fixed in the latest daily iso. Thank you @davidmhewitt!

davidmhewitt:
Great news! Thanks for reporting and testing!

PR #444, opened 2020-08-27
Set chosen language as default locale

davidmhewitt:
This _should_ make the chosen language be picked up by the guest (demo) session and hence fix (in theory) #442 .

I haven't been able to test on an installed system however as I can't get my guest session working at all.


danrabbit:
haha yup I can confirm that apt breaks without the "LANG="

PR #447, opened 2020-09-12
Reduce required ESP partition size to just under 100 MiB

### LINKED ISSUE #446: Odin installer minimum required EFI partition size is too large
The installer does not accept EFI partitions smaller than 512 MiB due to the constant in https://github.com/elementary/installer/blob/master/src/Views/PartitioningView.vala#L47

This is a request to set the constant to the minimum of 100 MB or 102400 sectors, assuming the sector size is 1024 bytes here.

Rationale for this issue report is that on my computers I first install Windows 10 which creates an ESP of 100 MB and then I proceed to install about 4 more instances of linux in a multi boot and the ESP size of 100 MB is more than enough.

I don't know the reason for the existing constant setting the minimum to 512 MB but I hope setting it to 100 MB won't be a problem?

vjr:
Fix #446 

This should enable people to install elementary OS as a dual boot with Windows 10 which creates a 100 MiB ESP, without requiring people to create another ESP to satisfy the old 256 MiB minimum size.


vjr:
If https://github.com/elementary/installer/pull/449 is approved then this PR can be discarded.

Issue #448, opened 2020-09-12
Installer support for non-512-byte sector sizes


vjr:
See https://github.com/pop-os/distinst/pull/181

Noticed this while working on https://github.com/elementary/installer/pull/447

vjr:
Looks like installer code already handles/uses the distinst disk `get_sector_size ()` value, except for the https://github.com/elementary/installer/blob/master/src/Objects/Mount.vala class.

This may only be relevant for the EFI ESP partition size check, in the (rare?) occasion the sector size is not 512 bytes, so this issue is probably low priority.

dr-Styki:
+1 With 4k sector size, a 256GB SSD is shown as a 32GB SSD
![With 4k sector size](https://user-images.githubusercontent.com/63053131/101897594-16a09080-3bab-11eb-9108-ecbe50f925cc.jpg)

Same SSD with 512b sector size
![With 512 sector size](https://user-images.githubusercontent.com/63053131/101897587-156f6380-3bab-11eb-8c33-8b199bd291a5.jpg)

patx:
This should be Priority: High

ryonakano:
Reopened since the issue in Distinst hasn't been fixed.

patx:
Crazy how this hasn't been addressed in years. It's why I stopped using elementary originally. I have an open issue on the distant project about this as well if anyone is interested.

PR #449, opened 2020-09-12
Consider sector size for ESP and reduce minimum size required

### LINKED ISSUE #446: Odin installer minimum required EFI partition size is too large

vjr:
Fixes #446 Addresses #448

If this is approved, supersedes #447 which won't need to be merged.


davidmhewitt:
After some experimenting relating to fixing installations performed with encryption enabled, it turns out distinst enforces a minimum ESP size (measured in sectors :scream: ), so if this was merged, it wouldn't work EFI sizes less than 256 MiB (on a 512 byte sector disk) anyway.

vjr:
I've filed https://github.com/pop-os/distinst/issues/239 in the hope that something can be worked out for us Windows users :-)

Issue #383, opened 2019-04-28
Smarter hostname


cassidyjames:
Split from #377. We should grab the model number as a default hostname, i.e. `Precision-5530` or `Galago-Pro`. Ubiquity does this as a user-changable default but I think we'd be fine just doing it behind the scenes.

btkostner:
As far as I know, there is no way to change it after install in the settings app. You can do it with terminal and it _should_ be simple, but some applications use it statically on first install or run. I personally think this should be overrideable on install.

cassidyjames:
@btkostner considering it's not even settable in the installer now, a smarter default would still be an improvement. However, Pop!_OS and other OSes don't offer a user-settable hostname and seem to work fine; changing it via Terminal is an easy one-time change for people who would like to dive into the internals of their computer.

Something I could see being considered that we've talked about is doing a pair of model and random hash so you have unique hostnames, even for a fleet of machines of the same model. Otherwise, I don't see a strong argument for making a user decide on an implementation detail.

cassidyjames:
Ubiquity does a couple of things:

1. Tries to get the model name from DMI, but there's a handful of special casing: https://git.launchpad.net/ubiquity/tree/ubiquity/misc.py#n618
2. If that fails, it uses `laptop-detect` and just appends `laptop` or `desktop`: https://git.launchpad.net/ubiquity/tree/ubiquity/plugins/ubi-usersetup.py#n77

I think trying DMI is worthwhile, though I'm less sure about the special casing; they obviously did it to work around weird cases but that seems annoying to maintain in the installer. I'd be more okay if it was its own library or something.

Instead of laptop-detect, `org.freedesktop.hostname1` has the `chassis` property which we could use.

btkostner:
@cassidyjames my point is, it's easy to set in the terminal, but there is a lot of software that uses that name on first run when generating configuration. Those files are static and are not updated with a simple terminal command.

cassidyjames:
@btkostner software like…? I've literally never come across that or had issues with changing my hostname. And regardless, that's still out of scope for this issue as there is _already_ no way to set a hostname in the installer. This doesn't change that, it just suggests making the _already hardcoded_ name a little smarter.

PR #453, opened 2020-09-16
Smarter hostname

meisenzahl:
Closes https://github.com/elementary/installer/issues/383


meisenzahl:
Do we want to use only a part of the `machine-id` for the hostname?

https://www.freedesktop.org/software/systemd/man/machine-id.html

tintou:
I don't really like relying on dmidecode for this, is there any alternative available (I'm pretty sure that dmidecode don't work in some contexts like VMs)

meisenzahl:
Dropped usage of `dmidecode` in favor of `/sys/devices/virtual/dmi/id/sys_vendor` and `/sys/devices/virtual/dmi/id/product_name`.

Current schema for hostname creation:

`<vendor>`-`<model>`-`<machine_id>[0:8]`

Fallback:

elementary-os-`<chassis>`-`<machine_id>[0:8]`

meisenzahl:
@cassidyjames Could you test if `/sys/devices/virtual/dmi/id/sys_vendor` and `/sys/devices/virtual/dmi/id/product_name` provide usable values on a Pinebook Pro?

davidmhewitt:
They likely won’t provide usable values since the pinebook pro doesn’t have DMI. But we’re also not using the installer there.

danrabbit (changes requested):
Hey thanks for working on this! I think since `get_hostname` is the only method that is accessed outside of this class, it's probably the only method that should be `public`. The others should be `private`

meisenzahl:
> Current schema for hostname creation:
> 
> `<vendor>`-`<model>`-`<machine_id>[0:8]`
> 
> Fallback:
> 
> elementary-os-`<chassis>`-`<machine_id>[0:8]`

What do you think about the hostname creation scheme?

I think the scheme of Ubiquity is built as follows:

`$USER`-`model`

danrabbit:
So for me, this makes my hostname `LENOVO-80MK-ee056906` by default, which is a little long and I'm not sure what the 80MK even is tbh. Machine ID might also be overkill.

`/sys/devices/virtual/dmi/id/product_version` actually yields `Lenovo YOGA 900-13ISK` which is still long, but much closer to what I would expect this computer to be called

I'm curious about what other folks are getting when they cat these files and if we can find something meaningful but relatively short

meisenzahl:
| Product name            | vendor                  | model                              | version                |
| ----------------------- | ----------------------- | ---------------------------------- | ---------------------- |
| Asus X201E              | `ASUSTeK COMPUTER INC.` | `X201EP`                           | `1.0       `           |
| Dell Latitude 7490      | `Dell Inc.`             | `Latitude 7490`                    |                        |
| HP 15-bw0xx             | `HP`                    | `HP Laptop 15-bw0xx`               | `Type1ProductConfigId` |
| HP Compaq Presario CQ60        | `Hewlett-Packard`       | `Compaq Presario CQ60 Notebook PC` | `F.31`                 |
| Lenovo ThinPad T14s     | `LENOVO`                | `20UHCTO1WW`                       | `ThinkPad T14s Gen 1`  |
| Lenovo ThinkStation S20 | `LENOVO`                | `4157GH7`                          | `ThinkStation XXXX`    |
| Raspberry Pi 4          |                         |                                    |                        |
| TREKSTOR Primebook P14  | `TREKSTOR`              | `Primebook P14`                    | `Default string`       |

The values in the table were queried on the hardware as follows:

```shell
vendor=`cat /sys/devices/virtual/dmi/id/sys_vendor`
model=`cat /sys/devices/virtual/dmi/id/product_name`
version=`cat /sys/devices/virtual/dmi/id/product_version`
```

The OEMs' implementation is a big mess 🙈

btkostner:
| Product name | vendor | model | version |
| --- | --- | --- | --- |
| Custom Built | `ASUS` | `All Series` | `System Version` |
| System76 Oryx Pro | `System76` | `Oryx Pro` | `oryp5` |

isantop:
Maybe unhelpful given the last response:

| Product name | vendor | model | version |
| --- | --- | --- | --- |
| System76 Oryx Pro | `System76` | `Oryx Pro` | `oryp4` |

danrabbit:
So it seems like for the data we have here so far that Model is usually better for everything except when vender is `LENOVO` lol. Would that be absolutely terrible to use version for Lenovo and Model for everyone else?

cassidyjames:
Throwing my hat in the ring since I have a decent amount of hardware around.

Product name             | vendor           | model            | version    | Note
------------             | ------           | -----            | -------    |
Dell Precision 5530      | `Dell Inc.`      | `Precision 5530` | ``         |
Slimbook Pro X (14")     | `SLIMBOOK`       | `PROX14`         | `Standard` | Slimbook often stylizes as "SLIMBOOK"
Pinebook Pro             | ``               | ``               | ``         | Fails because the `dmi` files don't exist
System76 Meerkat (meer1) | `System76, Inc.` | `Meerkat`        | `meer1`    |

@meisenzahl so to answer, no, Pinebook Pro does not have usable values here. :smile:

cassidyjames:
@meisenzahl @tintou @danrabbit the `dmidecode` logic in Ubiquity (and thus originally ported here) already has a bunch of fallbacks for different OEMs and whatnot, including Lenovo and VMs. Is it worth re-inventing the wheel here, or _does_ it make sense to build on that logic?

meisenzahl:
@tintou what is your opinion? Should I restore the ported logic from ubiquity without relying on `dmidecode` and read the values from `sys/devices/virtual/dmi/id/...`?

danrabbit:
@meisenzahl if you're using code derived from another work you need to update the copyright header in the file and make sure the license is compatible as well

meisenzahl:
According to https://launchpad.net/ubiquity, ubiquity is licensed under GPLv2. There are references to GPLv2 and GPLv3 in the code. The references to GPLv2 look like this: "either version 2 of the License, or (at your option) any later version." This means to me that GPLv3 as a later version is compatible with it. 

However, I am not a lawyer, so I don't know if we are compatible by using GPLv3.

src/Utils.vala
```diff
@@ -117,4 +118,129 @@ namespace Utils {
...
+        if (manufacturer.contains ("to be filled")) {
+            // Don't bother with products in development.
+            return null;
+        }
+
+        if (manufacturer.contains ("bochs") || manufacturer.contains ("vmware")) {
+            model = "virtual machine";
+            // VirtualBox sets an appropriate system-product-name.
+        } else {
+            string key = "system-product-name";
```
davidmhewitt, line 203: I don't get this. What is this `key` variable that's not used anywhere for?

meisenzahl: This was a remnant based on the call to `dmidecode`. I now read the corresponding file directly.

src/Utils.vala
```diff
@@ -117,4 +118,140 @@ namespace Utils {
...
+            }
+
+            if (model.down () == "not-available") {
+                return null;
+            }
+            if (model.down () == "To be filled by O.E.M.".down ()) {
+                return null;
+            }
+        } catch (RegexError e) {
+            print ("Error: %s\n", e.message);
```
davidmhewitt, line 245: Would be better to use `warning` here, and remove the `\n`. It would also be preferable to make the error message more clear so it means something in the installer logs (e.g. "Error cleaning up hostname strings: %s"). Probably also want to return null here too.

src/Utils.vala
```diff
@@ -117,4 +118,140 @@ namespace Utils {
...
+            var re = new Regex ("[^a-zA-Z0-9]+");
+            model = re.replace (model, model.length, 0, "-");
+            while (model[0] == '-') {
+                model = model.substring (1);
+            }
+            while (model[model.length - 1] == '-') {
+                model = model.substring (0, model.length - 1);
+            }
+
+            if (model.down () == "not-available") {
```
davidmhewitt, line 238: The general convention for hostnames is to keep them always lower case anyway. So I think it makes more sense to make sure all of the strings that can form `model` are already lower case, rather than always doing `model.down ()` for comparisons.

davidmhewitt (changes requested):

src/Utils.vala
meisenzahl: @danrabbit @cassidyjames do we want to lowercase hostnames? This would mean that, for example, `Star-LabTop-Mk-IV` would become `star-labtop-mk-iv`.

cassidyjames: I think if uppercase is valid, we should keep them; I do frequently use uppercase in my hostnames just fine 😉

davidmhewitt (approved):
Looks good to me, lets give it a try

danrabbit (approved):

PR #454, opened 2020-09-17
Application: Use Gtk.App inhibit

danrabbit:


davidmhewitt (commented):
Given that the installer session doesn't run under gala, the right interfaces for this to work may not necessarily be running. That said, I don't know how well the current inhibitor works either.

I'd really like to figure out a good testing flow for the installer.

zeebok (approved):

PR #458, opened 2020-11-03
Add search input to Language and Keyboard

alainm23:
Searching for a language or keyboard can take a while, a search entry might make this process faster.

image


danrabbit:
The reason we don't have a search input here is that you haven't configured keyboard layout yet, so we can't guarantee the input will be meaningful

PR #462, opened 2020-12-03
Set XDG_CURRENT_DESKTOP to Pantheon

davidmhewitt:
I haven't tested this but I believe it will work in conjunction with https://github.com/elementary/default-settings/pull/186/ to ensure the installer session presents itself with the correct XDG_CURRENT_DESKTOP name to pick up the correct GSettings overrides.


danrabbit (approved):
Nothing has blown up in the greeter so far. LFG

PR #465, opened 2020-12-10
ProgressView: Use LocaleHelper to set installation locale

### LINKED ISSUE #379: The installation fails with certain locales because of wrong locale code
Originally from https://github.com/pop-os/distinst/issues/183 (which was inappropriate to file there)

**Describe the bug**
The installation successfully finishes when I choose English (en_US), but it fails when I choose other languages, at least Japanese (ja_JP) or Korean (ko_KR).

**To Reproduce**
1. Select Japanese (which is shown as "日本語") in the language list
2. Choose clean install
3. The installation begins
4. The installation fails

**Expected behavior**
The installation should finish successfully even in Japanese or Korean.

**Screenshots**

This is the error message which is shown when I set the language to Japanese.

And this case is Korean.

 - OS: elementary OS Juno (Daily 20190426) on Virtual Box 6.0.6
 - elementary Installer 1.0+r1014+pkg13\~daily\~ubuntu5.0.1
 - Distinst 0.3.2\~1555431998\~18.04~3f51803

**Additional context**
As far as seeing the log, the reason why installation fails in Japanese seems to be related to wrong locale code:

```
…
(First part omitted)

INFO: running "chroot" "/tmp/distinst.bgLvf7fm49Ya" "locale-gen" "--purge" "ja_JA.UTF-8"
WARN: /bin/bash: warning: setlocale: LC_ALL: cannot change locale (ja_JA.UTF-8)
WARN: Error: 'ja_JA.UTF-8' is not a supported language or locale
ERROR: configuring chroot error: error generating locales: command failed with exit status: exit code: 1
```

The error message says "ja_JA", which does not exist locale code (the correct locale code for Japanese is "ja_JP"). I would say some logic to set locale does not work correctly.

davidmhewitt:
Use method introduced in #444 to set the locale on distinst.

This in theory supports locales where there is a different country code to the language code, currently we just uppercase the language code, which obviously doesn't work for locales like "ko_KR.UTF-8"

This is kind of an alternative to https://github.com/elementary/installer/pull/340 in that it in theory fixes the same thing, but doesn't use all of the distinst functionality that @tintou was questioning.

Once this is merged and we have working daily builds again, it'd be worth checking that this fixes #379


src/Views/ProgressView.vala
```diff
@@ -127,12 +127,18 @@ public class ProgressView : AbstractInstallerView {
...
-        } else {
-            config.lang = current_config.lang + "_" + current_config.lang.ascii_up () + ".UTF-8";
+            lang += "_" + current_config.country;
+        }
+
+        string? locale;
+        if (yield LocaleHelper.language2locale (lang, out locale)) {
+            if (locale != null) {
+                config.lang = locale;
+            }
```
tintou, line 141: Do we have some debugging/feedback available in case the language isn't found?

davidmhewitt: I guess this just silently falls back to en_US at the moment. I'm just trying to improve on the existing situation that selects invalid locales. I can add some debug messaging, but I don't think it gets logged anywhere during a real installation?

danrabbit (approved):

PR #467, opened 2021-01-13
Add workflow to build .deb packages

davidmhewitt:
When a PR is created, this builds the .deb packages for the installer and uploads them as an artifact to the GitHub actions CI run. This should make it slightly easier to test PRs as we can quickly download the built deb files and integrate them into an iso.


vjr:
What will happen when a draft PR is created? I would think the intent is this should not get triggered in this case?

davidmhewitt:
I think it would still run, but it doesn't really matter since it doesn't block anything, and maybe you do want a .deb package of your draft PR to test.

.github/workflows/build-deb.yml
```diff
@@ -0,0 +1,44 @@
...
+          apt update
+          apt install -y git
+
+      - name: Checkout main
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+
+      - name: Checkout debian packaging folder
+        run: git checkout origin/deb-packaging -- debian
```
btkostner, line 28: This might be able to be replaced with another `actions/checkout@v2` call, but considering this works as is right now, no need to update it.

btkostner (approved):

.github/workflows/build-deb.yml
davidmhewitt: Hm, this is a trick to checkout the `debian` folder from the `deb-packaging` branch into the currently checked out commit, essentially merging them. I don't think `actions/checkout@v2` has anything that can do that? Otherwise, we'd need to check out the repository again into a different folder and then copy the `debian` folder across.

btkostner: There is a way to set it to clone in a subdir and a specific branch, just not sure on the merge part. But anyway, no need to worry about it. If this works, I'm happy.

Issue #457, opened 2020-09-28
Installation fails to boot on UEFI systems if disk encryption was enabled


marbetschar:
**Describe the bug**

I downloaded elementary OS Daily from 27th of Sept 2020 and did a bare metal installation on my Dell XPS 13. The installation went smooth without any issue with everything set to default - but after it completed and the machine did a restart, Odin fails to boot claiming it is not able to find a kernel (right after the Grub Boot Menu).

Then I re-installed without Disk Encryption and then Odin was able to boot.

**Expected behavior**

Being able to boot after a bare metal installation with enabled disk encryption :)

![Screenshot from 2020-09-28 08 22 47](https://user-images.githubusercontent.com/392542/94397296-d4e37a80-0163-11eb-8bc6-5d52b6107471.png)

mick-wilson:
I experienced something similar, however I wasn't sure if it was this issue or https://github.com/elementary/installer/issues/438

In any case, I worked around it by 

* Going to the "advanced" disk configuration part of the wizard
* opening gparted and manually wiping the disk entirely
* Backing out of "advanced" and selecting "clean install" again. 

Once I had the disk wiped manually, even the encrypted "clean install" worked successfully.

tyuio9:
I guess, it might be connected with Grub support for LUKS2 (Odin uses LUKS2). https://git.savannah.gnu.org/cgit/grub.git/commit/?id=365e0cc3e7e44151c14dd29514c2f870b49f9755

meisenzahl:
I am not able to reproduce this issue with `elementaryos-6.0-daily.20201207`

@marbetschar can you confirm this?

fabricio7p:
i'm also having trouble installing any build into my 2019 XPS 13 9300.
Install fails with
`ERROR: configuring chroot error: error installing packages`

Right after extracting files, i can see in the logs it tries to run chroot apt-get command and fails there.
Tried everything default, with internet.. No luck

marbetschar:
@meisenzahl, as @fabricio7p said the current iso builds do have some installation issues. Will happily re-install on bare metal once the issues are resolved - cause I need a running machine to prepare for the exams ;)

fabricio7p:
Just a note, my disk is not encrypted. Should i open another issue?
Do you have any ideas on why its failing on a chroot command?

disaster123:
Same for me. Grub does not find the uuid after enabling encryption. Used image from 20210107

davidmhewitt:
[...]

PR #470, opened 2021-01-22
Create a /boot mount point if we're using encryption

davidmhewitt:
This is the first part of the fixes for #457 

I've tested that this fixes boot failures on UEFI systems installed using the "Clean Install" option with encryption enabled. It works by creating an extra unencrypted `/boot` mount point for the kernel and initramfs files.

This will not help with a custom install option where you're installing into an existing encrypted volume.

### DIFF
--- a/src/Views/ProgressView.vala
+++ b/src/Views/ProgressView.vala
@@ -191,11 +191,20 @@ public class ProgressView : AbstractInstallerView {
             value = 0
         };
 
-        var boot_sector = Distinst.Sector () {
+        // 256 MiB is the minimum distinst ESP partition size, so this is 256 MiB in MB plus a bit
+        // extra for safety
+        var efi_sector = Distinst.Sector() {
             flag = Distinst.SectorKind.MEGABYTE,
-            value = 512
+            value = 278
         };
 
+        // 512MB /boot partition that's created if we're doing encryption
+        var boot_sector = Distinst.Sector () {
+            flag = Distinst.SectorKind.MEGABYTE,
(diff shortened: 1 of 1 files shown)


src/Views/ProgressView.vala
```diff
@@ -191,11 +191,20 @@ public class ProgressView : AbstractInstallerView {
             value = 0
         };
 
+        // 256 MiB is the minimum distinst ESP partition size, so this is 256 MiB in MB plus a bit
```
meisenzahl, line 194: Just to get this right:

256 MiB ≈ 268,435 MB

So ~10 MB for safety?

meisenzahl (commented):
I just have a question to understand it better. Otherwise looks good to me.

src/Views/ProgressView.vala
davidmhewitt: Approximately yes, since we're working with disk sectors, there will be some amount on the start of the drive we can't use too. So in reality, it'll probably be more like 6-8MB for safety.

I did set it at 256MiB and then distinst failed it because of the reserved sectors on the start of the drive, so I just added 10MB or so 😃

meisenzahl: Thanks for clarification

meisenzahl (approved):
LGTM

davidmhewitt:
@cassidyjames @danrabbit This is the PR I was talking about during standup that fixes disk encryption for clean installs. Do we just want to merge it and see how it goes?

PR #478, opened 2021-02-08
Split into GUI and Daemon

davidmhewitt:
Requires deb packaging branch: #479 

This looks like quite a scary diff, but it's largely just moving things out of the current installer process into a new daemon and then adding the necessary structures to communicate with it over DBus.

I've tested BIOS and UEFI installs with this branch, with both custom and default partitioning and all is working well.

The only thing this changes functionally is some of the hangs in the GUI where Distinst was doing stuff on the GUI thread. So things feel a little more responsive, especially around disk selection and partitioning now.

This doesn't yet make the necessary changes to have the GUI run as a standard user. I'll follow up with that in a separate branch as this one is getting kind of long.

It would be good to get this merged sooner rather than later so we can test for regressions in the daily isos and unblock the rest of the work around wrong theming, polkit stuff in the live session, keyboard layouts etc.

### DIFF
--- a/common/DBusStructures.vala
+++ b/common/DBusStructures.vala
@@ -58,12 +58,12 @@ public enum InstallerDaemon.MountFlags {
 }
 
 public struct InstallerDaemon.Mount {
-    public string partition_path;
-    public string parent_disk;
-    public string mount_point;
-    public uint64 sectors;
-    public Distinst.FileSystem filesystem;
-    public MountFlags flags;
+    string partition_path;
+    string parent_disk;
+    string mount_point;
+    uint64 sectors;
+    Distinst.FileSystem filesystem;
+    MountFlags flags;
 
     public bool is_valid_boot_mount () {
         return filesystem == Distinst.FileSystem.FAT16
@@ -86,7 +86,7 @@ public struct InstallerDaemon.Mount {
 }
 
 public struct InstallerDaemon.LuksCredentials {
-    public string device;
-    public string pv;
-    public string password;
+    string device;
+    string pv;
+    string password;
 }
--- a/common/DBusStructures.vala
+++ b/common/DBusStructures.vala
@@ -34,5 +34,40 @@ public struct InstallerDaemon.InstallConfig {
 }
 
 public struct InstallerDaemon.PartitionConfig {
-    string placeholder;
+    Mount[] mounts;
+}
+
+[Flags]
+public enum InstallerDaemon.MountFlags {
+    FORMAT = 1,
+    LVM = 2,
+    LVM_ON_LUKS = 4
+}
+
+public struct InstallerDaemon.Mount {
+    public string partition_path;
+    public string parent_disk;
+    public string mount_point;
+    public uint64 sectors;
+    public Distinst.FileSystem filesystem;
+    public MountFlags flags;
+
+    public bool is_valid_boot_mount () {
+        return filesystem == Distinst.FileSystem.FAT16
+            || filesystem == Distinst.FileSystem.FAT32;
+    }
+
+    public bool is_valid_root_mount () {
+        return filesystem != Distinst.FileSystem.FAT16
+            && filesystem != Distinst.FileSystem.FAT32
+            && filesystem != Distinst.FileSystem.NTFS;
+    }
+
+    public bool is_lvm () {
+        return MountFlags.LVM in flags;
+    }
+
+    public bool should_format () {
+        return MountFlags.FORMAT in flags;
+    }
 }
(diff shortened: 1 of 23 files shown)


common/DBusStructures.vala
```diff
@@ -0,0 +1,92 @@
+/*
+ * Copyright (c) 2021 elementary, Inc.
```
tintou, line 2:
```suggestion
 * Copyright 2021 elementary, Inc.
```

src/Views/CheckView.vala
```diff
@@ -340,6 +357,6 @@ public class Installer.CheckView : AbstractInstallerView {
 }
 
 [DBus (name = "org.freedesktop.UPower")]
-public interface UPower : GLib.Object {
+public interface UPower : DBusProxy {
```
tintou, line 360: It is better to not do this, the DBusProxy is automatically created by Vala

src/Views/CheckView.vala
```diff
@@ -145,7 +162,7 @@ public class Installer.CheckView : AbstractInstallerView {
             try {
                 upower = Bus.get_proxy_sync (BusType.SYSTEM, "org.freedesktop.UPower", "/org/freedesktop/UPower", GLib.DBusProxyFlags.GET_INVALIDATED_PROPERTIES);
 
-                (upower as DBusProxy).g_properties_changed.connect ((changed, invalid) => {
+                upower.g_properties_changed.connect ((changed, invalid) => {
```
tintou, line 165: See comment here-under

src/Views/PartitioningView.vala
```diff
@@ -115,40 +137,45 @@ public class Installer.PartitioningView : AbstractInstallerView {
...
+            return;
+        }
 
-            string path = Utils.string_from_utf8 (disk.get_device_path ());
+        foreach (unowned InstallerDaemon.Disk disk in disks.physical_disks) {
+            var sector_size = disk.sector_size;
+            var size = disk.sectors * sector_size;
 
-            string model = Utils.string_from_utf8 (disk.get_model ());
+            string path = disk.device_path;
```
tintou, line 158:
```suggestion
            unowned string path = disk.device_path;
```

src/Views/PartitioningView.vala
```diff
@@ -173,24 +200,23 @@ public class Installer.PartitioningView : AbstractInstallerView {
...
-        var sector_size = disk.get_sector_size ();
-        var size = disk.get_sectors () * sector_size;
-
-        string path = Utils.string_from_utf8 (disk.get_device_path ());
+    private void add_logical_disk (InstallerDaemon.Disk disk) {
+        var sector_size = disk.sector_size;
+        var size = disk.sectors * sector_size;
 
-        string model = Utils.string_from_utf8 (disk.get_model ());
+        string path = disk.device_path;
```
tintou, line 210:
```suggestion
        unowned string path = disk.device_path;
```

src/Views/PartitioningView.vala
```diff
@@ -230,42 +256,16 @@ public class Installer.PartitioningView : AbstractInstallerView {
...
-        });
-
-        switch (result) {
-            case 0:
-                unowned Distinst.LvmDevice disk = disks.get_logical_device (pv);
+    private void on_partition_decrypted (InstallerDaemon.LuksCredentials credentials) {
+        luks.add (credentials);
+        Daemon.get_default ().get_logical_device.begin (credentials.pv, (obj, res) => {
+            try {
+                var disk = Daemon.get_default ().get_logical_device.end (res);
```
tintou, line 263:
```suggestion
                var disk = ((Daemon) obj).get_logical_device.end (res);
```

src/Widgets/DecryptMenu.vala
```diff
@@ -118,6 +122,52 @@ public class Installer.DecryptMenu: Gtk.Popover {
         pass_entry.grab_focus_without_selecting ();
     }
 
+    private async void decrypt (string pv) {
+        string password = pass_entry.get_text ();
```
tintou, line 126:
```suggestion
        unowned string password = pass_entry.get_text ();
```

tintou:
don't forget to update the POTFILES

davidmhewitt:
@tintou Thanks for the review! I've not updated POTFILES because there aren't any translatable strings in the new files. Would you still advise to add them to POTFILES?

src/meson.build
```diff
@@ -27,7 +28,8 @@ vala_files = [
     'Widgets/PartitionBar.vala',
     'Widgets/PartitionMenu.vala',
     'Widgets/Terminal.vala',
-    'Widgets/VariantWidget.vala'
+    'Widgets/VariantWidget.vala',
+    join_paths(meson.source_root(), 'common', 'DBusStructures.vala'),
```
tintou, line 32:
```suggestion
    join_paths('..', 'common', 'DBusStructures.vala'),
```
`meson.source_root()` is deprecated

tintou (approved):
Let's merge this!

PR #486, opened 2021-02-20
Use greeter to switch to demo mode

### LINKED ISSUE #456: Demo mode gparted unusable due to lacking permissions
The new demo mode seems to lack the permissions to access gparted, which is nesessary if someone wants to preformat their storage, or recover their data

### LINKED ISSUE #468: Demo mode asks for password when connecting to WiFi
**Describe the bug**

When using the daily build of elementary OS 6 Early Access (as of January 16, 2021) and using the demo mode, it does not connect to the WiFi networks. When trying to connect to a protected network, it asks for the password for the `elementary` user (which is unknown to me).

**To Reproduce**

1. Boot into the daily build iso file of elementary OS 6
2. Try to connect to a WiFi network
3. See the password prompt

**Expected behavior**

The temporary guest user in the demo mode should be able to connect to WiFi.

- OS: elementary OS 6 Early Access (Linux 5.4.0-62-generic)

**Additional context**

Somewhat related to #439.

davidmhewitt:
Requires deb packaging branch: https://github.com/elementary/installer/pull/487

Fixes #468
Fixes #456

Allows switching to the demo mode using the actual live user, instead of a guest user. Resolving any permissions issues.

To be used in conjunction with:
https://github.com/elementary/greeter/pull/509


src/Utils.vala
```diff
@@ -60,12 +60,15 @@ namespace Utils {
         if (Installer.App.test_mode) {
             critical (_("Test mode switch user"));
         } else {
-            get_system_instance ();
+            // This touches the file `/var/lib/lightdm/demo-mode`, which signals to the greeter that the next session it launches
+            // should be the live (demo) session. If this file doesn't exist, it just relaunches the installer session
+            Installer.Daemon.get_default ().trigger_demo_mode.begin ();
```
tintou, line 65: Should we make sure that this got correctly executed (and that we have no racing) before doing anything and log the error with `debug`

tintou (approved):

PR #495, opened 2021-02-27
Rework success/reboot strings slightly

### LINKED ISSUE #305: Use ngettext when possible
Because l18n and stuff
https://github.com/elementary/installer/blob/5ed4b356a0e09ec5407165851143cce910bed553/src/Views/SuccessView.vala#L42

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/56981707-use-ngettext-when-possible?utm_campaign=plugin&utm_content=tracker%2F46189108&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F46189108&utm_medium=issues&utm_source=github).
</bountysource-plugin>

davidmhewitt:
Fixes #305 

People keep mistranslating the original `"Your device will automatically restart to %s in %i seconds."` string by reordering the two printf arguments and not knowing how to specify the order. This is causing crashes.

See https://github.com/elementary/installer/commit/088d83806d0e52ec64b8a9a2bdaf2b5a402c6b95 (where I fix the issue for German) and then https://github.com/elementary/installer/commit/523bf6f567f580e44ddb2d23d76c52f814f72987 where someone immediately reverts it back to its broken, crashing state.

Additionally, we weren't properly handling the fact we have a plural number of seconds here which may need to be handled differently in different languages.

This changes the meaning slightly in that it will say something like

"elementary OS 6 was successfully installed. Your device will automatically restart in 30 seconds. After restarting you can set up a new user, or you can shut down and set up a new user later."

Instead of:

"Your device will automatically restart to elementary OS 6 in 30 seconds. After restarting you can set up a new user, or you can shut down and set up a new user later."

However, the fact it will be much more difficult for translators to cause crashes and the fact we use ngettext properly here is probably worth it.


src/Views/SuccessView.vala
```diff
@@ -31,7 +31,12 @@ public class SuccessView : AbstractInstallerView {
         primary_label.get_style_context ().add_class (Granite.STYLE_CLASS_H2_LABEL);
 
         var secondary_label = new Gtk.Label (
-            _("Your device will automatically restart to %s in %i seconds.").printf (Utils.get_pretty_name (), RESTART_TIMEOUT) + " " +
+            _("%s was successfully installed.").printf (Utils.get_pretty_name ()) + " " +
```
cassidyjames, line 34: Looks like we use “has been installed” for apps in AppCenter and Sideload, so I would match that here. Plus we don’t need to specify “successfully” when it is implied.


```suggestion
            _("%s has been installed.").printf (Utils.get_pretty_name ()) + " " +
```

cassidyjames (changes requested):
Tiny copy tweak otherwise lgtm

danrabbit (approved):
:rocket:

PR #498, opened 2021-03-02
MainWindow: Ask WM to CENTER_ALWAYS

### LINKED ISSUE #496: Re-center if display resolution changes
I'm seeing an issue where I'm guessing the display resolution has changed after the installer window was created, because it is way off center in the top left corner.

We should problem listen for resolution changes and make sure we recenter the window

danrabbit:
Fixes #496 

Remove deletable while we're here since we don't pack window controls anymore

### DIFF
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -35,12 +35,12 @@ public class Installer.MainWindow : Hdy.Window {
 
     public MainWindow () {
         Object (
-            deletable: false,
             height_request: 700,
             icon_name: "system-os-installer",
             resizable: false,
             title: _("Install %s").printf (Utils.get_pretty_name ()),
-            width_request: 950
+            width_request: 950,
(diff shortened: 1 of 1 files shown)


cassidyjames (changes requested):
There's no reason to delete `deletable`; it's unrelated to centering the window, and controls more than just window controls—i.e. closing the window with Alt+F4 depending on the window manager. Otherwise, yeah, this asks the window manager to keep it centered which is what we want.

src/MainWindow.vala
```diff
@@ -35,12 +35,12 @@ public class Installer.MainWindow : Hdy.Window {
 
     public MainWindow () {
         Object (
-            deletable: false,
             height_request: 700,
```
cassidyjames, line 38:
```suggestion
            deletable: false,
            height_request: 700,
```

cassidyjames (approved):

PR #500, opened 2021-03-05
EncryptView: Warn about keyboard

### LINKED ISSUE #499: Warn about keyboards when encrypting
This was brought recently when testing a mini desktop; Plymouth doesn’t support Bluetooth or touch input, so when encrypting, we should warn that you'll need a built-in or USB keyboard instead of Bluetooth or touch.

cassidyjames:
Fixes #499.

- Replaces the performance warning with a keyboard warning.
- Updates code style on the lines touched.
- Pulls out magic number for label length into a const.
- Removes the now-unused performance icon.
- Bumps copyright while we're here.

Screenshot from 2021-03-05 13-05-01

We should provide the larger size of symbolic icon to be used here. https://github.com/elementary/icons/issues/1016

Only outstanding question for me is: do we need to specifically call out touch and Bluetooth? I struggled doing that without making it too wordy.


danrabbit (approved):
I think this works :+1: I'm not sure touch etc need to be specifically called out

Issue #503, opened 2021-03-24
[Feature request] Make preseed/kickstart/unattended install possible


swirly:
I'm in a team that deploy linux in large academic environnement. As of today, we're using essentially Debian because of the real ease of automated install : you can preseed the installer with even an http remote file ! There are others types of methods to do this, named kickstart or unattended. Even Microsoft Windows installer can be fully automated !

We'd like to deploy elementary OS, but we need that the installation can be fully automated. When you have just one person and one day to deploy 100 to 250 desktop computers for pupils, you can't seat in front of each computer : pxe boot, preseeded installation and voila !

The addition of this use case would be a real killer feature. Indeed for us it would be necessary :)

TIA for your consideration

swirly:
Thank you. If I can be of any help in the design process, I will gladly participate

meisenzahl:
I have implemented a minimal prototype ^

It works based on a JSON configuration file.

@swirly Could you please explain a bit more how your current setup works with Debian. Do you have to modify the ISO file or how does it work?

swirly:
In debian, there is a pressed file that can be combined with the iso without modifying the iso.
If you use a USB key, the preseed file can be on the usb key
But, far more interessant, the preseed file can be accessed through http with a kernel parameter at startup. For pxe installation, you just have to :
 - boot on pxe
 - in pxe, select via menu "debian installation"
 - and in the pxe configuration, the preseed file is mentionned. 

It can even be created server side on the fly, so you can have different setup for different machines.

The complete description of this brilliant mechanism is here : 
[Debian preseed](https://wiki.debian.org/DebianInstaller/Preseed)

possible preseed source are file, floppy, http, https and tftp

The preseed format is really a flat file, simple to understand and write
[example preseed file](https://www.debian.org/releases/stable/example-preseed.txt)

Note that the preseed file allow to automate installation not only of the installer but also give pre-recorded answer for all debian packages.

Hope this helps

Conan-Kudo:
This is also true with kickstart (which can be used with Debian systems with `kickseed`). The `inst.ks` kernel parameter can be read by the installer to load the kickstart.

swirly:
Almost two years since report. Any news on this ?

swirly:
Three year and a half since opening. 

Should the issue be marked as won't-fix or is there any roadmap on this ?

Actually, installing elementary OS on an academic part is the best way to promote the OS. What you use in school is a reference for future. But for now, this installation is not possible without automatic deployement.

[...]

PR #504, opened 2021-04-02
Make preseed/kickstart/unattended install possible

meisenzahl:
Closes https://github.com/elementary/installer/issues/503

Simple solution to load the configuration from a JSON file.


meisenzahl:
I have tried this a few times now. Works so far quite well. I start the installer in test mode with `--test`. Thereupon at the end in the `SuccessView` a JSON configuration file is written to `/tmp/io.elementary.installer.json`. After that I start the installer on the command line with `io.elementary.installer -f /tmp/io.elementary.installer.json`. Then the configuration is read and it immediately jumps to the `ProgressView` and installs elementary OS as described in the read configuration.

PR #510, opened 2021-04-30
Make titles consistent

MrPotatoBobx:
Fixes https://github.com/elementary/installer/issues/307
 
 I have changed most options to start with a verb but i do not know how to word the "custom(advanced)" option

PR #525, opened 2021-05-05
TryInstallView: Rewrite as radiobuttons, code style

### LINKED ISSUE #515: Confusing headers in the UI
**Describe the bug**
I think that the new UI is a bit confusing. The description below refers to a screenshot that I posted in the _Screenshots_ section.

I was sure that the "Try or Install" header on the left is actually a button, and that there are in total 4 options to choose from:
 - the "Try or Install" header
 - the 3 other buttons on the left

I think that the header lacks some "contrast" to distinguish it from the actual content that the user is supposed to interact with.

Another issue that can be seen in the screenshot are the horizontal margins - they're way too small.

**To Reproduce**
Ru the installer for eOS6

**Expected behavior**
The UI should make it clear what is the header, and what are the buttons.

**Screenshots**
image

**Smartphone (please complete the following information):**

**Additional context**

danrabbit:
Fixes #515 

Screenshot from 2021-05-05 09 46 12

* Remove redundant label in the title area. We don't use a label like this in the other preceding views
* Use radio buttons to make it obvious where the selection is, and very clear what the current selection is
* Incorporate some string changes from #510 
* Lots of code style and organizational cleanups

### DIFF
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -36,11 +36,11 @@ public class Installer.MainWindow : Hdy.Window {
     public MainWindow () {
         Object (
             deletable: false,
-            height_request: 700,
+            default_height: 700,
+            default_width: 950,
             icon_name: "system-os-installer",
             resizable: false,
             title: _("Install %s").printf (Utils.get_pretty_name ()),
-            width_request: 950,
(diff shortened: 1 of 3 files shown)


cassidyjames (approved):
Nice, this is much cleaner and clearer. :+1:

PR #529, opened 2021-05-05
DiskView: Rewrite as RadioButtons, code style

danrabbit:
Screenshot from 2021-05-05 11 32 30

* Make selection more obvious by using radio buttons
* Use modern code style and style constants
* Remove unnecessary expands
* Move description label to the right, to be consistent with other views title area
* Add left/right margins to the view for consistency


cassidyjames:
Hm, the vertical alignment looks a little off here; in master, the right side is vertically-centered, but now it feels too far up. Not sure what's going on there.

cassidyjames (approved):

PR #531, opened 2021-05-05
SuccessView: Redesign, code style

danrabbit:
Screenshot from 2021-05-05 12 20 43

* Put title in the left side like other views
* Make "OS has been installed" a primary label
* Update the secondary label as we count down
* Reduce max width chars so we aren't forcing the window larger
* Remove application.css. This appears to be a workaround for a fixed stylesheet issue
* Fix margins
* Code style


cassidyjames (changes requested):
We definitely need to wrap the seconds in tabular number markup otherwise, that string jiggles all over the place.

cassidyjames (approved):
dope

PR #532, opened 2021-05-05
 TryInstallView: stop view from vexpanding the window

danrabbit:
* Don't wrap title label, this preserves extra space
* Add max_width_chars to subtitle to force the wrap


cassidyjames (approved):
Does the thing.

PR #535, opened 2021-05-05
ProgressView: Add translator comments

comradekingu:


cassidyjames:
These don't really need ellipsis? It's describing the current step, and a percent is shown next to it.

comradekingu:
True enough, but the convention remains consistent this way.
Since the context isn't carried over, errors in translation are avoided.
"Configuring the system" is likely to be ~"This is how to set up the system" for example.

cassidyjames:
@comradekingu the correct thing may be to include a translator comment above that string in code if it could be more clear, e.g.

```vala
///TRANSLATORS: Used to communicate the current step of the installer back-end
```

Or something

cassidyjames (approved):

PR #540, opened 2021-05-06
MainWindow: Use default height/width

danrabbit:
height and width request enforce a minimum size, so the window can't be made smaller. Use default height/width instead. This probably doesn't stop the window from being created larger than the screen, but it's a step :shrug: 

These values are much smaller because height/width request apparently includes the shadow or something? But the window itself is the same size afaict

### DIFF
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -36,11 +36,11 @@ public class Installer.MainWindow : Hdy.Window {
     public MainWindow () {
         Object (
             deletable: false,
-            height_request: 700,
+            default_height: 600,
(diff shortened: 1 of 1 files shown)


cassidyjames:
I know we've avoided default height/width in the past due to the shadow issue. This seems to change the size of the window by just a few pixels (so not a big deal), but doesn't actually make it resizable or anything. Is that intended?

danrabbit:
Yeah that's correct it doesn't make it resizable from the user side, it just changes the property we're setting from "minimum size" to "default size"

PR #586, opened 2021-08-07
Installer encryption copy updates

orta:
I think this cleans up the wording. I opted for the minimal changes,to be fair though, I think "added protection" doesn't sell the underlaying reason well though. Perhaps?

> "Encrypting this device's drive is required to ensure the safety of your information. Encrypting your drive means:"

Edit: I've spotted the "Don't encrypt" button so "required" is pretty much wrong in my PR. What about:

> "Encrypting this device's drive ensures the safety of your information. We recommend enabling drive encryption which means:"


ryonakano (changes requested):
Thanks for your contribution! Would you revert changes against `.po` files? They're automatically updated by GitHub Actions, so you don't need to touch them. Also they mess up the diff 😅

cassidyjames:
@orta the "if required" copy was very intentional, as some organizations require end users to use full disk encryption; this helps the user know why they might want to. I agree it could be worded better, though.

cassidyjames (changes requested):
@orta the "if required" copy was very intentional, as some organizations require end users to use full disk encryption; this helps the user know why they might want to. I agree it could be worded better, though. We also have to be really careful not to make promises we can't guarantee; "added protection" is accurate, but we can't really "ensure" anything.

src/Views/EncryptView.vala
```diff
@@ -52,7 +52,7 @@ public class EncryptView : AbstractInstallerView {
         };
         title_label.get_style_context ().add_class (Granite.STYLE_CLASS_H2_LABEL);
 
-        var details_label = new Gtk.Label (_("Encrypt this device's drive if required for added protection, but be sure you understand:")) {
+        var details_label = new Gtk.Label (_("Encrypting this device's drive ensures the safety of your information. We recommend enabling drive encryption which means:")) {
```
orta, line 55: What about this instead:

```suggestion
        var details_label = new Gtk.Label (_("Encrypting this device's drive improves the safety of your information. We recommend enabling drive encryption which means:")) {
```

cassidyjames: It still feels weird to say we recommend it, but then the default is not to enable it (for good reason, namely the localization, input method, and sync issues). I'm not totally against rewording this, but again, the original copy was carefully iterated several times for very specific reasons.

PR #621, opened 2021-10-08
Handle disk sector size

### LINKED ISSUE #448: Installer support for non-512-byte sector sizes

### LINKED ISSUE #620: Installer shows incorrect partition sizes, cannot install due to perceived ESP partition size
### What Happened?

Attempting to install elementary OS 6 fails on my 2019 MacBook Pro, seemingly due to not being able to correctly read partition sizes on the disk. Attempting to install on this machine via the default "Erase Disk and Install" option fails with a "Installing elementary OS 6 Odin failed, possibly due to a hardware error." message, and attempting a "Custom Install" does not allow me to continue.

### Steps to Reproduce

1. Boot into macOS Recovery to disable Secure Boot, booting from external partitions, etc.; delete existing partitions from disk
2. Boot into elementary OS 6 installer and attempt to "Erase Disk and Install"
   1. Select the internal 1 TB NVME drive ("Apple SSD AP1024M, /dev/nvme0n1 1.0 TB")
   2. Choose "Don't Encrypt" when prompted (though it doesn't matter — I've tried encrypting before and the failure is the same)
   3. Hit "Could Not Install" error
      <details><summary>Screenshot</summary></details>
      I'm not sure whether the "pointer in FFI is null" error is unrelated, correlated, or causative, but the last error produced is related to ESP partition being too small
3. Go back with "Try Installing Again", this time selecting a "Custom Install"
   1. "Modify Partitions" to create a 512 MB EFI partition, 16 GB swap partition, 64 GB root partition, and remaining home partition 
      <details><summary>Screenshot</summary></details>
   2. See that installer is incorrectly picking up on partition sizes (67.1 MB for EFI, 2.1 GB for swap, 8.6 GB for root, 114.3 GB for home, 875.5 GB _unused_)
      <details><summary>Screenshot</summary></details>
   3. Attempt to select the partitions anyway (installation can't begin because EFI partition is too small)
      <details><summary>Screenshot</summary></details>
 
I've confirmed that booting into Demo Mode on the same drive shows the right partition sizes in GParted, as does booting into a GParted live USB.

### Expected Behavior

* "Erase and Install" should have succeeded at blowing away anything on disk and installing the OS
* The partition selector should also show the same partition sizes as "Modify Partitions"

### OS Version

6.x (Odin) [`elementaryos-6.0-stable.20211005`]

### Software Version

### Log Output

I re-ran the above steps as-is (including recreating all partitions) this morning (the day after submission) and grabbed `journalctl` output from demo mode: [journal.log](https://github.com/elementary/installer/files/7311648/journal.log)

### Hardware Info

This is on a 2019 MBP with a 1 TB internal drive and SIP and Secure Boot previously disabled.

vjr:
Fixes #620 

Requires https://github.com/pop-os/distinst/pull/266


vjr:
Fixes #448

src/Widgets/PartitionBar.vala
```diff
@@ -89,6 +91,10 @@ public class Installer.PartitionBar : Gtk.EventBox {
         return end - start;
     }
 
+    public uint64 get_sector_size () {
```
davidmhewitt, line 94: This doesn't seem to be used anywhere, so could be removed?

davidmhewitt (dismissed):
I've imported a version of distinst with the associated fixes into the elementary PPAs, and the code looks pretty good here, just one small thing to change and then I'll test this out.

I don't like passing this `sector_size` through all of these layers of construction since it's just a property of the disk, not of every partition and mount, but that isn't a problem with your code, it's jut the original architecture of this, so we can sort that later.

src/Widgets/PartitionBar.vala
vjr: done!

vjr:
@davidmhewitt  @elementary/desktop-developers  I've created PR https://github.com/pop-os/distinst/pull/278 and along with this PR it appears to succeed installation for the 3 test cases I tried in QEMU:

1. Custom install - disk partitioned with a 500 MB EFI and the rest formated ext4 mounted on / (root)
2. Erase and install without encryption
3. Erase and install with encryption

Note that I say "appears to succeed" because the installer succeeds and partitions appear to be created fine, I've not rebooted into the installed image in QEMU to verify it works post install - would appreciate if anyone else would be able to test and comment.

In the mean time I'll try to more fully test in the coming days and report my findings here.

benoitaguesse:
> Has there been any progress on this? I'm still hitting the 4096k block size issue in 24.04 LTS when trying to install Pop!_OS on a 2017 MacBook Pro

Me too, any hope this will be adrdessed soon ?

PR #624, opened 2021-11-16
CheckView: Fix wrong gettext syntax

ryonakano:
This fixes the string always showing in English regardless of the selected language


danrabbit (approved):
Nice catch

PR #631, opened 2021-12-17
Daemon: double encrypted boot size

danrabbit:
Daemon got a big ol' boot


cassidyjames (approved):
I like big `/boot` and I cannot lie

PR #633, opened 2021-12-17
Explicitly select disk

### LINKED ISSUE #250: "Connect to a Power Source" screen should disapear on ac power

13r0ck:
## Before
The user can select a drive without the `Erase and Install` being sensitive. It is not clear enough.

https://user-images.githubusercontent.com/58987761/146617494-12a694ce-b976-4b54-b5ad-0a635b10af79.mp4

## After
User is required to select the drive. Either with the mouse or `Space`

https://user-images.githubusercontent.com/58987761/146617506-69d2eedd-1fb4-4345-b426-74f7394dd6db.mp4

This also fixes #250


13r0ck:
Wrong repo. Sorry!

PR #635, opened 2021-12-20
MainWindow: replace stack with Deck

danirabbit:
This assumes that It's always okay to go back except on progress view and success view. We may want to either disable navigation or make corrections on some pages if this isn't the case.

* Remove views when we get back to the language view or at the try/install view instead of every time we navigate
* Fixes check view navigation being backwards
* Views are more agnostic of their position and use navigate (forward/back)
* Check view is shown for custom partitioning


lenemter (commented):
Transition from CheckView to DisckView or PartitioningView is not so smooth because of removing CheckView. As a solution we can just keep it, or remove only when the next view is opened.

src/MainWindow.vala
```diff
@@ -291,25 +241,16 @@ public class Installer.MainWindow : Hdy.Window {
...
-
-        error_view = new ErrorView (log);
-        stack.add (error_view);
-        stack.visible_child = error_view;
-
-        error_view.previous_view = disk_view;
+        var error_view = new ErrorView (log);
+        deck.add (error_view);
+        deck.visible_child = error_view;
+        deck.can_swipe_back = true;
```
lenemter, line 253: This should be false? Otherwise it is possible to swipe back to progress view and further back.

lenemter (changes requested):
When pressing "Try Installing Again" button in ErrorView it navigates back to a progress view, where you can't do anything.

danirabbit:
@lenemter ah yep. It looks like previously it would send you back to the disks view, but I'm now having it send you back to the try/install view since you could have gotten there from the custom partitioning path as well

src/MainWindow.vala
```diff
@@ -291,25 +241,19 @@ public class Installer.MainWindow : Hdy.Window {
...
-        stack.add (error_view);
-        stack.visible_child = error_view;
+        var error_view = new ErrorView (log);
+        deck.add (error_view);
+        deck.visible_child = error_view;
 
-        error_view.previous_view = disk_view;
+        error_view.retry_install.connect (() => {
+            deck.visible_child = try_install_view;
+        });
```
lenemter, line 256:
```suggestion
        error_view.retry_install.connect (() => {
            deck.visible_child = try_install_view;
             deck.can_swipe_back = true;
        });
```

danirabbit: Good call!

lenemter (approved):

PR #648, opened 2022-04-30
MainWindow: throw an infobar on battery

danrabbit:
Currently, if you go from powered to battery we throw you from wherever you were into the CheckView to warn you about installing on battery power. This branch throws an infobar covering the beginning of the action area instead.

The reasoning behind the placement here is to making sure you can still complete the language view if this is shown on startup. Another alternative would be waiting to throw the infobar until after you've completed the language view and then we could show it in the more traditional top aligned location

Screenshot from 2022-04-30 11 15 37

### DIFF
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -23,6 +23,7 @@ public interface UPower : GLib.Object {
 }
 
 public class Installer.MainWindow : Hdy.Window {
+    private Gtk.Label infobar_label;
     private Gtk.Stack stack;
 
     private LanguageView language_view;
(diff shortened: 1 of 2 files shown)


tintou (changes requested):
You need to listen to the selected language and reset the label on language selection so that it is translated in the language the user has chosen

danrabbit:
@tintou would setting it again like this work or do I need to destroy and rebuild the label?

tintou:
@danrabbit no this won't work, the `_()` function has to be called again so you need to regenerate the string

danrabbit:
@tintou okay I believe it's fixed now

tintou (approved):

PR #649, opened 2022-05-24
CheckView: rewrite as a single page

### LINKED ISSUE #639: Minimize check view
### Problem

It's kind of a pain for each check to be its own view. We don't really need a whole page to say, for example, plug in a laptop.

### Proposal

We should probably have a single check view page that lists each concern on it in one or two lines

### Prior Art (Optional)

_No response_

danirabbit:
Fixes #639 

Screenshot from 2022-05-24 13 32 37

Only add whichever lines are relevant:

Screenshot from 2022-05-24 13 34 16


jeremypw (approved):
Makes sense and works as expected.  Its a bit odd for the private class to have same name as the parent, especially as it is no longer an independent view.  Not blocking for this PR thought.

PR #657, opened 2022-10-28
Ship own app icon

danirabbit:
Don't rely on a non-fdo icon name


micahilbery (approved):
:green_circle: :arrow_down: :+1:

PR #665, opened 2022-11-28
Views use 128px icons

danirabbit:
Use 128px icons like Initial Setup.

Create a custom icon for check view, use computer-fail for VM, and change label. `computer-fail` here is a little too pessimistic for this whole view I think. This is more of like a "proceed at your own risk" thing than an outright fail. I also changed the label here to "Before Installing" since this page covers more than system requirements

Screenshot from 2022-12-16 14 13 35


src/Views/CheckView.vala
```diff
@@ -111,11 +112,11 @@ public class Installer.CheckView : AbstractInstallerView {
             critical ("Couldn't read apt sources: %s", e.message);
         }
 
-        if (get_vm ()) {
+        if (get_vm () || Installer.App.test_mode) {
             message_box.add (vm_view);
         }
 
-        if (!minimum_specs) {
+        if (!minimum_specs || Installer.App.test_mode) {
```
ryonakano, line 119: Is this change meant to added for testing this branch? This might look a little strange when someone run the installer in the test mode on real machines.

danirabbit: Removed!

ryonakano (approved):

PR #668, opened 2022-11-28
CheckView: warn about daily builds

### LINKED ISSUE #507: Show a warning if the installation image is beta software
## TD;LR 
---
Since public beta will release soon(https://github.com/elementary/builds/pull/63), I would recommend adding some kind of warning that informs the user that the installation image that they are using is beta release software.

I think it should be shown as soon as the image is booted up (it may scare some users though at it might seem like an error)

or, we can show it when the end user clicks any of the options in the main installer (demo mode, clean install and advanced install)

Here is a concept I made in Krita (excuse my drawing skills)
popup concept

danirabbit:
Fixes #507 

Screenshot from 2022-11-28 11 04 18

### DIFF
--- a/src/Views/CheckView.vala
+++ b/src/Views/CheckView.vala
@@ -108,7 +108,7 @@ public class Installer.CheckView : AbstractInstallerView {
         var apt_sources = File.new_for_path ("/etc/apt/sources.list.d/elementary.list");
         try {
             var @is = apt_sources.read ();
-            var dis  = new DataInputStream (@is);
+            var dis = new DataInputStream (@is);
 
             if ("daily" in dis.read_line ()) {
                 message_box.add (beta_view);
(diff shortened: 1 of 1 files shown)


src/Views/CheckView.vala
```diff
@@ -99,6 +105,18 @@ public class Installer.CheckView : AbstractInstallerView {
             minimum_specs = false;
         }
 
+        var apt_sources = File.new_for_path ("/etc/apt/sources.list.d/elementary.list");
```
meisenzahl, line 108: In onboarding we read is info: https://github.com/elementary/onboarding/blob/master/src/MainWindow.vala#L53

Did you already try this here? Would be nice if we could use the same logic.

meisenzahl (changes requested):

src/Views/CheckView.vala
danirabbit: Yeah, this seems flawed though because lsb info comes from the os-patches repo which is applied to both daily and stable, so this stops working as soon as we set LSB to not say "early access"

meisenzahl (approved):

PR #669, opened 2022-11-28
CheckView: show when messagebox has children

danirabbit:
Show this view whenever there are messages in the list. Better than checking individual variables like this because we don't have to update it every time this view changes


meisenzahl (approved):
Nice cleanup

PR #670, opened 2022-11-28
CheckView: remove minimum space check

danirabbit:
We already check disk size in the DiskView so I'm not sure why we'd do that twice


zeebok (approved):
Certainly is odd to have it on both. Code looks good to me!

PR #675, opened 2022-12-10
Make ESP optional for custom installs

### LINKED ISSUE #662: Make EFI ESP optional during custom install
### Problem

Currently configuring an ESP is required before the "Erase and install" button is enabled in PartitionView to start the custom installation.

I have KDE neon as my other install and use the bootloader from there but the elementary (v7 horus daily) bootloader "takes precedence" over neon even if I install it before or after installing neon.

When I try to repair by either running "sudo grub-install" or efibootmgr to try and restore the neon bootloader it sticks to elementary - not sure what is going on here but perhaps something to do with reusing ubuntu's grub binaries?

### Proposal

Make the requirement to mount the ESP optional - only during custom install mode - so that those with dual boot can preserve their bootloader of choice.

### Prior Art (Optional)

IIRC the ubuntu ubiquity installer has/had an option to not install the bootloader - either in the UI or command line `-b` option.

vjr:
Fixes #662


src/Views/PartitioningView.vala
```diff
@@ -225,7 +225,6 @@ public class Installer.PartitioningView : AbstractInstallerView {
             case Distinst.PartitionTable.MSDOS:
                 break;
             case Distinst.PartitionTable.GPT:
-                required |= Defined.EFI;
```
zeebok, line 228: Is there a way to detect if the system is in EFI mode to enforce this requirement when it makes sense?

vjr: This PR is to exclude the requirement (make it optional) for an EFI system partition (ESP) for custom install mode in any case.

I presume anyone choosing to do a custom install would be familiar with the requirement for an ESP and will add one with the partitioning tool if they are, say, installing only elementary OS on their machine in custom mode.

zeebok: Yeah good point. For some reason when I wrote this in my head this was also for regular installs as well

zeebok (approved):

vjr:
@elementary/desktop-developers build failing - CI/development target with `"Dependency libdistinst-dev not found"` ... can someone help fix this (please commit directly to this PR) unless it's just a matter of waiting for the package to become available?

PR #677, opened 2023-01-08
actions/checkout@v3

meisenzahl:


.github/workflows/gettext.yml
```diff
@@ -6,9 +6,9 @@ jobs:
   gettext_template:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v1
-    - uses: elementary/actions/gettext-template@master
-      env:
-        GIT_USER_TOKEN: "${{ secrets.GIT_USER_TOKEN }}"
-        GIT_USER_NAME: "elementaryBot"
-        GIT_USER_EMAIL: "builds@elementary.io"
+      - uses: actions/checkout@v3
```
danirabbit, line 9: We've had problems in the past with newer checkout breaking pushing for the gettext action, so we should verify that this works somewhere before changing it

danirabbit (approved):

PR #701, opened 2023-04-09
Fix #617 : Raise DBUS_TIMEOUT_MSEC value in src/Helpers/InstallerDaemon.vala

### LINKED ISSUE #617: Installer not showing disk or partitions, can't install.
## What happened

The new installer does not seem to recognize the ssd disk on a 2016 Macbook Pro. Trying to "erase and install" gives a "no enough space" message (and shows no disk), and the "custom install" option shows no disk. Gparted gives a "Invalid argument during the preparation for reading at /dev/nvme0n2" (I think that must be some minimal size recovery mac-related memory). The message can be ignored, and the partitions can be made in nvme0n1, but the installer does not recognize it. There's some snapshots [here](https://elementaryos.stackexchange.com/questions/28643/elementary-os-odin-installer-not-showing-partitions-on-macbook-pro-2016).

Note that I've checked that Ubuntu 20.04, 21.04 and elementary 5.1 recognize the disk correctly, so it seems that this could be the new installer's fault.

## Expected Behavior
The disk should be recognized to perform the installation.

## Steps to Reproduce
1. Launch the ISO
2. Select the disk
3. Install

## Platform Information
* elementary OS 6 Odin
* Macbook Pro 2016

aerocyber:
- Fixes #617 

- A higher value for `DBUS_TIMEOUT_MSEC` did manage to fix the issue while testing from the most recent live cd from the official website.


src/Helpers/InstallerDaemon.vala
```diff
@@ -17,7 +17,11 @@
 
 public class Installer.Daemon {
     // Wait up to 60 seconds for DBus calls to timeout. Some of the Distinst disk probe operations seem to take around 30 seconds
-    private const int DBUS_TIMEOUT_MSEC = 60 * 1000;
+    // private const int DBUS_TIMEOUT_MSEC = 60 * 1000;
```
lenemter, line 20: Remove this redundant please

src/Helpers/InstallerDaemon.vala
```diff
@@ -17,7 +17,11 @@
...
+    
+    private const int DBUS_TIMEOUT_MSEC = 60 * 100000000;
```
lenemter, line 22: 1. Isn't 69 days a bit too long? I would think it should be something like 3 minutes at max
2. This value is supposed to be `seconds` * `milliseconds`, so change it to match that please

src/Helpers/InstallerDaemon.vala
```diff
@@ -17,7 +17,11 @@
 
 public class Installer.Daemon {
     // Wait up to 60 seconds for DBus calls to timeout. Some of the Distinst disk probe operations seem to take around 30 seconds
-    private const int DBUS_TIMEOUT_MSEC = 60 * 1000;
+    // private const int DBUS_TIMEOUT_MSEC = 60 * 1000;
+    
+    private const int DBUS_TIMEOUT_MSEC = 60 * 100000000; 
+    // The original value of 60 * 1000 may not be sufficient time for getting partition information in older systems.
+    // A higher value is thus assigned. Of course, it increases the time required for the installer to recognize partitions.
```
lenemter, line 24: Instead of writing all of this, update the comment above at line 19 :)

lenemter (changes requested):

src/Helpers/InstallerDaemon.vala
aerocyber: 1. Yeah, I did not meant it to be that long.
2. Thank you, have corrected it.

src/Helpers/InstallerDaemon.vala
aerocyber: Of course!

aerocyber:
Can you please guide me with the linting error?

lenemter:
@aerocyber There is a trailing whitespace at line 21. Probably it's inserted automatically by your text editor when you press enter.

lenemter (approved):
Looks great! Thank you for this one. :rocket:

aerocyber:
Thank you for your help and time!

PR #707, opened 2023-06-24
Increase DBUS_TIMEOUT_MSEC variable

aerocyber:
This is done so that on some systems, particularly the older ones, the installer fails to continue in `Custom Install` due to not showing ALL partitions.

### DIFF
--- a/src/Helpers/InstallerDaemon.vala
+++ b/src/Helpers/InstallerDaemon.vala
@@ -16,10 +16,10 @@
  */
 
 public class Installer.Daemon {
-    // Wait up to 300 seconds for DBus calls to timeout. Some of the Distinst disk probe operations seem to take around 30 seconds.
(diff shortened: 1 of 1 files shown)


aerocyber:
Note: This is a follow-up for pr #701 as I noticed in my recent re-installation of elementary, it fails even at the previous value of 300 seconds (5 minutes). This one did work for me. Should it be even higher value if a really old system is taken into account? Won't probing the disk stop after it's successful or will it run for the entire 10 minute duration?

lenemter:
> Won't probing the disk stop after it's successful or will it run for the entire 10 minute duration?

It will stop stop when it succeeds. This timeout sets maximum time that installer will wait for probing.

lenemter (approved):

lenemter:
For some reason development-target CI fails and I can't merge this. @danirabbit Can you merge this, please?

danirabbit:
@lenemter temporarily removed the requirement for `development-target`

PR #711, opened 2023-07-21
Prompt for screen reader shortcut after a timeout

### LINKED ISSUE #471: Automatically prompt for screen reader after a timeout
A person with low vision pointed out that a feature they heavily appreciate in macOS is that the screen reader is automatically enabled after a certain amount of inactivity at the installer. Turning on the screen reader is the absolute most important first step for them to be able to interact with the computer. Without it, they pointed out that they would have to rely on a sighted person to set up their computer for them, which is a huge hindrance for them.

danirabbit:
Fixes #471

Read out a screen reader prompt after 10 seconds if the user hasn't navigated in this time


danirabbit:
I'm guessing this doens't need to be localized because you'd need to either be able to see or activate the screen reader before selecting a language

src/MainWindow.vala
```diff
@@ -102,6 +107,27 @@ public class Installer.MainWindow : Hdy.Window {
         battery_infobar.response.connect (() => {
             battery_infobar.revealed = false;
         });
+
+        var mediakeys_settings = new Settings ("org.gnome.settings-daemon.plugins.media-keys");
+
+        orca_timeout_id = Timeout.add_seconds (3, () => {
```
zeebok, line 113: You said 10 secs in the description but it is 3 here

danirabbit: Whoops. Thank you lol

src/MainWindow.vala
```diff
@@ -78,6 +78,11 @@ public class Installer.MainWindow : Hdy.Window {
         add (overlay);
 
         language_view.next_step.connect (() => {
+            // Don't prompt for screen reader if we're able to navigate without it
+            if (orca_timeout_id != 0) {
+                Source.remove (orca_timeout_id);
```
zeebok, line 83: This remove call seems to kick a critical error to output for the ID not existing if you wait for the timeout to occur. I checked and the ID is definitely non-zero so I am not sure what is happening there, or if it is something we should be concerned about.

danirabbit: I think I fixed it. Thanks for the attention to detail!

zeebok: Huh. Interesting that assigning the timeout ID to Source.REMOVE isn't taking care of that.

zeebok: But that fixed it!

zeebok (approved):
Looks good. I have a feeling the timeout length may end up needing future tweaking. There is no search on this first page so those whom need to scroll a bunch to find their language and layout may end up surprised by the voice.

PR #712, opened 2023-07-21
Control: Add espeak to depends

danirabbit:
Not sure if should be depends instead? Goes with #711


zeebok:
To me this feels important enough to depend on it as eOS seems to put a strong priority on a11y

zeebok (approved):

PR #719, opened 2023-07-25
Add an option to install proprietary drivers

### LINKED ISSUE #418: No option to install proprietary drivers
In the previous ubiquity installer, there was a checkbox to install proprietary drivers and codecs. In 5.1, this option would install the `bcmwl-kernel-source` package (for my Broadcom adapter) into the target system so that I had working WiFi when I booted into the installed system.

The new installer no longer has this option and have to dig out a USB ethernet adapter or remember the arcane commands to install the driver off the live media to get my WiFi working after install.

lenemter:
Fixes #418
Closes #666

Screenshot from 2023-07-26 09 53 18

### DIFF
--- a/src/Views/DriversView.vala
+++ b/src/Views/DriversView.vala
@@ -38,7 +38,7 @@
 
         var warning_image = new Gtk.Image.from_icon_name ("security-low-symbolic", LARGE_TOOLBAR);
 
-        var warning_label = new Gtk.Label (_("Proprietary drivers contain closed-source components. You may prefer not to install them due to hardware compatibility issues.")) {
+        var warning_label = new Gtk.Label (_("Proprietary drivers contain private code that can't be reviewed. Security and other updates are dependent on the driver vendor.")) {
             hexpand = true,
             max_width_chars = 1, // Make Gtk wrap, but not expand the window
             wrap = true,
--- a/src/Views/DriversView.vala
+++ b/src/Views/DriversView.vala
@@ -19,7 +19,7 @@
     public signal void next_step ();
 
     construct {
-        var image = new Gtk.Image.from_icon_name ("application-x-executable", Gtk.IconSize.DIALOG) {
+        var image = new Gtk.Image.from_icon_name ("application-x-firmware", Gtk.IconSize.INVALID) {
             pixel_size = 128,
             valign = Gtk.Align.END
         };
--- a/src/Views/DriversView.vala
+++ b/src/Views/DriversView.vala
@@ -29,7 +29,7 @@
         };
         title_label.get_style_context ().add_class (Granite.STYLE_CLASS_H2_LABEL);
 
-        var description_label = new Gtk.Label (_("Proprietary drivers are required for some devices, such as Wi-Fi adapters or graphics cards, to work properly.")) {
+        var description_label = new Gtk.Label (_("Some devices, such as NVIDIA® graphics or Broadcom® Wi-Fi adapters, may not function properly without proprietary drivers. Most devices do not require additional drivers.")) {
             max_width_chars = 1, // Make Gtk wrap, but not expand the window
             wrap = true,
             xalign = 0
--- a/src/Views/DriversView.vala
+++ b/src/Views/DriversView.vala
@@ -47,7 +47,7 @@
 
         var legal_image = new Gtk.Image.from_icon_name ("text-x-copying-symbolic", LARGE_TOOLBAR);
 
-        var legal_label = new Gtk.Label (_("These drivers come with their own licensing terms. For that reason, we cannot redistribute them freely.")) {
+        var legal_label = new Gtk.Label (_("Proprietary drivers are subject to their own license terms. By installing, you agree to their terms of use.")) {
             hexpand = true,
             max_width_chars = 1, // Make Gtk wrap, but not expand the window
             wrap = true,
--- a/src/Views/DriversView.vala
+++ b/src/Views/DriversView.vala
@@ -56,7 +56,7 @@
 
         var install_later_image = new Gtk.Image.from_icon_name ("system-software-install-symbolic", LARGE_TOOLBAR);
 
-        var install_later_label = new Gtk.Label (_("You can always install them later through AppCenter. But in this case, you may encounter non-working hardware on the first boot.")) {
+        var install_later_label = new Gtk.Label (_("Proprietary drivers can be installed later through AppCenter, but this will require a working Internet connection.")) {
             hexpand = true,
             max_width_chars = 1, // Make Gtk wrap, but not expand the window
             wrap = true,
--- a/src/Views/DriversView.vala
+++ b/src/Views/DriversView.vala
@@ -24,7 +24,7 @@
             valign = Gtk.Align.END
         };
 
-        var title_label = new Gtk.Label (_("Install Drivers")) {
+        var title_label = new Gtk.Label (_("Additional Drivers")) {
             valign = Gtk.Align.START
         };
         title_label.get_style_context ().add_class (Granite.STYLE_CLASS_H2_LABEL);
--- a/src/Views/DriversView.vala
+++ b/src/Views/DriversView.vala
@@ -69,15 +69,7 @@
             xalign = 0
         };
 
-        var drivers_label = new Gtk.Label (_("Install Proprietary Drivers:"));
-
-        var drivers_switch = new Gtk.Switch ();
-
-        var drivers_box = new Gtk.Box (HORIZONTAL, 12) {
-            halign = CENTER
-        };
-        drivers_box.add (drivers_label);
-        drivers_box.add (drivers_switch);
+        var drivers_check = new Gtk.CheckButton.with_label (_("Include third-party proprietary drivers when installing"));
 
         var message_grid = new Gtk.Grid () {
             valign = CENTER,
@@ -91,7 +83,7 @@
         message_grid.attach (legal_label, 1, 2);
         message_grid.attach (install_later_image, 0, 3);
         message_grid.attach (install_later_label, 1, 3);
-        message_grid.attach (drivers_box, 0, 4, 2, 1);
+        message_grid.attach (drivers_check, 0, 4, 2, 1);
 
         content_area.column_homogeneous = true;
         content_area.margin_start = content_area.margin_end = 12;
@@ -110,9 +102,9 @@
         action_area.add (back_button);
         action_area.add (next_button);
 
-        drivers_switch.notify["active"].connect (() => {
+        drivers_check.toggled.connect (() => {
             unowned var configuration = Configuration.get_default ();
-            configuration.install_drivers = drivers_switch.active;
+            configuration.install_drivers = drivers_check.active;
         });
 
         show_all ();
--- a/src/Views/DriversView.vala
+++ b/src/Views/DriversView.vala
@@ -48,6 +48,8 @@
         };
 
         var legal_image = new Gtk.Image.from_icon_name ("text-x-copying-symbolic", LARGE_TOOLBAR);
+        legal_image.get_style_context ().add_class ("accent");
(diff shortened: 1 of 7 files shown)


danirabbit:
We probably need some more explanation of what “proprietary” means. Why someone wouldn’t want to do this. Legal implications or possible technical issues etc. basically why is this is an option and not just something that is always done

danirabbit:
Maybe also we can be more specific about which devices this covers?

Another question might be if someone chooses not to, will they be able to install these drivers later? If so, how?

src/Views/DriversView.vala
```diff
@@ -0,0 +1,114 @@
...
+        var warning_label = new Gtk.Label (_("Proprietary drivers contain closed-source components. You may prefer not to install them due to hardware compatibility issues.")) {
+            hexpand = true,
+            max_width_chars = 1, // Make Gtk wrap, but not expand the window
+            wrap = true,
+            xalign = 0
+        };
+
+        var legal_image = new Gtk.Image.from_icon_name ("text-x-copying-symbolic", LARGE_TOOLBAR);
+
+        var legal_label = new Gtk.Label (_("These drivers come with their own licensing terms. For that reason, we cannot redistribute them freely.")) {
```
danirabbit, line 50: I think redistribution is an issue with codecs, but I'm not sure that it is with drivers. I'm having a hard time finding out what these licenses actually are though and how to review them

```suggestion
        var legal_label = new Gtk.Label (_("Proprietary drivers are subject to their own license terms. By installing, you agree to their terms of use.")) {
```

src/Views/DriversView.vala
```diff
@@ -0,0 +1,114 @@
...
+            pixel_size = 128,
+            valign = Gtk.Align.END
+        };
+
+        var title_label = new Gtk.Label (_("Install Drivers")) {
+            valign = Gtk.Align.START
+        };
+        title_label.get_style_context ().add_class (Granite.STYLE_CLASS_H2_LABEL);
+
+        var description_label = new Gtk.Label (_("Proprietary drivers are required for some devices, such as Wi-Fi adapters or graphics cards, to work properly.")) {
```
danirabbit, line 32: Maybe a little more specific so folks know who this is and isn't for

```suggestion
        var description_label = new Gtk.Label (_("Some devices, such as NVIDIA® graphics or Broadcom® Wi-Fi adapters, may not function properly without proprietary drivers. Most devices do not require additional drivers.")) {
```

src/Views/DriversView.vala
```diff
@@ -0,0 +1,114 @@
...
+            max_width_chars = 1, // Make Gtk wrap, but not expand the window
+            wrap = true,
+            xalign = 0
+        };
+        description_label.get_style_context ().add_class (Granite.STYLE_CLASS_H3_LABEL);
+
+        var warning_image = new Gtk.Image.from_icon_name ("security-low-symbolic", LARGE_TOOLBAR);
+
+        var warning_label = new Gtk.Label (_("Proprietary drivers contain closed-source components. You may prefer not to install them due to hardware compatibility issues.")) {
```
danirabbit, line 41: Mimic the Ubuntu copy on this

```suggestion
        var warning_label = new Gtk.Label (_("Proprietary drivers contain private code that can't be reviewed. Security and other updates are dependent on the driver vendor.")) {
```

src/Views/DriversView.vala
```diff
@@ -0,0 +1,114 @@
...
+        title_label.get_style_context ().add_class (Granite.STYLE_CLASS_H2_LABEL);
+
+        var description_label = new Gtk.Label (_("Proprietary drivers are required for some devices, such as Wi-Fi adapters or graphics cards, to work properly.")) {
+            max_width_chars = 1, // Make Gtk wrap, but not expand the window
+            wrap = true,
+            xalign = 0
+        };
+        description_label.get_style_context ().add_class (Granite.STYLE_CLASS_H3_LABEL);
+
+        var warning_image = new Gtk.Image.from_icon_name ("security-low-symbolic", LARGE_TOOLBAR);
```
danirabbit, line 39: Let's get some color in here :)

```suggestion
        var warning_image = new Gtk.Image.from_icon_name ("security-low-symbolic", LARGE_TOOLBAR);
        warning_image.get_style_context ().add_class ("accent");
        warning_image.get_style_context ().add_class ("yellow");
```

src/Views/DriversView.vala
```diff
@@ -0,0 +1,114 @@
...
+
+        var warning_label = new Gtk.Label (_("Proprietary drivers contain closed-source components. You may prefer not to install them due to hardware compatibility issues.")) {
+            hexpand = true,
+            max_width_chars = 1, // Make Gtk wrap, but not expand the window
+            wrap = true,
+            xalign = 0
+        };
+
+        var legal_image = new Gtk.Image.from_icon_name ("text-x-copying-symbolic", LARGE_TOOLBAR);
```
danirabbit, line 48:
```suggestion
        var legal_image = new Gtk.Image.from_icon_name ("text-x-copying-symbolic", LARGE_TOOLBAR);
        legal_image.get_style_context ().add_class ("accent");
        legal_image.get_style_context ().add_class ("slate");
```

src/Views/DriversView.vala
```diff
@@ -0,0 +1,114 @@
...
+
+        var legal_label = new Gtk.Label (_("These drivers come with their own licensing terms. For that reason, we cannot redistribute them freely.")) {
+            hexpand = true,
+            max_width_chars = 1, // Make Gtk wrap, but not expand the window
+            wrap = true,
+            xalign = 0
+        };
+
+        var install_later_image = new Gtk.Image.from_icon_name ("system-software-install-symbolic", LARGE_TOOLBAR);
```
danirabbit, line 57:
```suggestion
        var install_later_image = new Gtk.Image.from_icon_name ("system-software-install-symbolic", LARGE_TOOLBAR);
        install_later_image.get_style_context ().add_class ("accent");
        install_later_image.get_style_context ().add_class ("purple");
```

danirabbit (changes requested):
Screenshot from 2023-07-26 09 31 50

Just a few suggestions for copy etc. Looking really good!

src/Views/DriversView.vala
```diff
@@ -0,0 +1,114 @@
...
+        next_button.get_style_context ().add_class (Gtk.STYLE_CLASS_DESTRUCTIVE_ACTION);
+        next_button.clicked.connect (() => next_step ());
+
+        action_area.add (back_button);
+        action_area.add (next_button);
+
+        drivers_switch.notify["active"].connect (() => {
+            unowned var configuration = Configuration.get_default ();
+            configuration.install_drivers = drivers_switch.active;
+        });
```
danirabbit, line 110: I think since this is something we're including as part of a list (of things to install later) and not a feature that we're toggling on and off in real time, a checkbox is probably the more appropriate widget. Also, Ubiquity seems to hammer in the third-party-ness of the drivers here, which seems like a good idea:

Screenshot from 2023-07-26 09 47 22

```suggestion
        var drivers_check = new Gtk.CheckButton.with_label (_("Include third-party proprietary drivers when installing"));

        var message_grid = new Gtk.Grid () {
            valign = CENTER,
            row_spacing = 32,
            column_spacing = 12
        };
        message_grid.attach (description_label, 0, 0, 2, 1);
        message_grid.attach (warning_image, 0, 1);
        message_grid.attach (warning_label, 1, 1);
        message_grid.attach (legal_image, 0, 2);
        message_grid.attach (legal_label, 1, 2);
        message_grid.attach (install_later_image, 0, 3);
        message_grid.attach (install_later_label, 1, 3);
        message_grid.attach (drivers_check, 0, 4, 2, 1);

        content_area.column_homogeneous = true;
        content_area.margin_start = content_area.margin_end = 12;
        content_area.valign = Gtk.Align.CENTER;
        content_area.attach (image, 0, 0);
        content_area.attach (title_label, 0, 1);
        content_area.attach (message_grid, 1, 0, 1, 2);

        var back_button = new Gtk.Button.with_label (_("Back"));
        back_button.clicked.connect (() => ((Gtk.Stack) get_parent ()).visible_child = previous_view);

        var next_button = new Gtk.Button.with_label (_("Erase and Install"));
        next_button.get_style_context ().add_class (Gtk.STYLE_CLASS_DESTRUCTIVE_ACTION);
        next_button.clicked.connect (() => next_step ());

        action_area.add (back_button);
        action_area.add (next_button);

        drivers_check.toggled.connect (() => {
            unowned var configuration = Configuration.get_default ();
            configuration.install_drivers = drivers_check.active;
        });
```

src/Views/DriversView.vala
```diff
@@ -0,0 +1,114 @@
...
+ public class DriversView : AbstractInstallerView {
+    public signal void next_step ();
+
+    construct {
+        var image = new Gtk.Image.from_icon_name ("application-x-firmware", Gtk.IconSize.INVALID) {
+            pixel_size = 128,
+            valign = Gtk.Align.END
+        };
+
+        var title_label = new Gtk.Label (_("Install Drivers")) {
```
danirabbit, line 27: Just to make it clear that we're not installing drivers immediately in this step, but making a choice about installing them later. Going with the language "additional" here to imply that the system already comes with drivers, but there are more available optionally

```suggestion
        var title_label = new Gtk.Label (_("Additional Drivers")) {
```

src/Views/DriversView.vala
lenemter: > Most devices do not require additional drivers

A may be wrong, but Nvidia GPUs are more common than Amd or Intel and therefore most devices actually require these drivers.

davidmhewitt:
Just to add a bit more context here in case it's useful:

This will use `ubuntu-drivers` during the install process, and AFAIK the only hardware that really supports is NVIDIA GPUs and Broadcom WiFi cards. I think there might be some special cases in there for Ubuntu certified OEM hardware (specific Dell models etc), but for the most part, it's just NVIDIA and Broadcom.

Distros do have the rights to redistribute NVIDIA and Broadcom drivers, and in fact, we do distribute the Broadcom drivers on the iso already. The licensing is just a bit more restrictive in terms of "no reverse engineering" / "don't modify the drivers" / "don't do things we don't like with them" etc...

But if the user doesn't agree to those terms, they don't get working WiFi and have to throw their WiFi card in the bin and buy an Intel one. Soooooo... :laughing: 

I think this whole thing is just making it clear we're complying with NVIDIA's and Broadcom's terms by telling the users that they're actually agreeing to some corporate EULA by using and installing the drivers. So as long as the copy is at least as strong as Ubuntu's, I think we're good.

src/Views/DriversView.vala
danirabbit: Ah by "devices" here I mean individual hardware devices, not computers. But I just poked around and [according to this random and not well researched source](https://www.hardwaretimes.com/amds-dgpu-market-share-4-1-nvidia-20-intels-igpu-share-at-62-q3-2021-overall-gpu-market-share/) integrated graphics is almost 80% of the market

danirabbit:
@davidmhewitt Do you wanna read through again? I think we might have addressed all of those while you were typing :sweat_smile:

davidmhewitt:
> @davidmhewitt Do you wanna read through again? I think we might have addressed all of those while you were typing sweat_smile

Yeah, seems good!

src/Views/DriversView.vala
lenemter: Oh makes sense then

Btw thanks for taking over the design of this, I couldn't have done it properly :)

danirabbit (approved):
I'm happy with this in terms of design. Code seems straightforward, but while we have @davidmhewitt's attention maybe he can give a glance since I'm not familiar with distinst etc :)

lenemter:
@danirabbit I added #666 here for quick testing, should I remove it?

danirabbit:
@lenemter nah could just add “closes” to the description here 👍

PR #722, opened 2023-07-27
Fix size of icon in Encrypt View

### LINKED ISSUE #718: "Enable Drive Encription" icon is smaller than the others
### What Happened?

image

image

### Steps to Reproduce

1. Open Installer
2. Naviagate to encription view

### Expected Behavior

Big icon

### OS Version

7.x (Horus)

### Software Version

Compiled from git

### Log Output

_No response_

### Hardware Info

_No response_

zeebok:
Fixes #718

### DIFF
--- a/src/Views/EncryptView.vala
+++ b/src/Views/EncryptView.vala
@@ -32,7 +32,10 @@ public class EncryptView : AbstractInstallerView {
     }
 
     construct {
-        var image = new Gtk.Image.from_icon_name ("drive-harddisk", Gtk.IconSize.DIALOG);
+        var image = new Gtk.Image.from_icon_name ("drive-harddisk", Gtk.IconSize.DIALOG) {
+            pixel_size = 128,
+            valign = Gtk.Align.END
+        };
 
         var overlay_image = new Gtk.Image.from_icon_name ("security-high", Gtk.IconSize.DND) {
             halign = Gtk.Align.END,
--- a/src/Views/EncryptView.vala
+++ b/src/Views/EncryptView.vala
@@ -33,8 +33,7 @@ public class EncryptView : AbstractInstallerView {
 
     construct {
         var image = new Gtk.Image.from_icon_name ("drive-harddisk", Gtk.IconSize.INVALID) {
-            pixel_size = 128,
-            valign = Gtk.Align.END
+            pixel_size = 128
         };
 
         var overlay_image = new Gtk.Image.from_icon_name ("security-high", Gtk.IconSize.INVALID) {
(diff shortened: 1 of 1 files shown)


src/Views/EncryptView.vala
```diff
@@ -32,7 +32,10 @@ public class EncryptView : AbstractInstallerView {
     }
 
     construct {
-        var image = new Gtk.Image.from_icon_name ("drive-harddisk", Gtk.IconSize.DIALOG);
+        var image = new Gtk.Image.from_icon_name ("drive-harddisk", Gtk.IconSize.DIALOG) {
```
danirabbit, line 35: Should be INVALID since it's a different size than the stock sizes

```suggestion
        var image = new Gtk.Image.from_icon_name ("drive-harddisk", INVALID) {
```

src/Views/EncryptView.vala
```diff
@@ -32,7 +32,10 @@ public class EncryptView : AbstractInstallerView {
...
+            pixel_size = 128,
+            valign = Gtk.Align.END
+        };
 
         var overlay_image = new Gtk.Image.from_icon_name ("security-high", Gtk.IconSize.DND) {
```
danirabbit, line 40: We probably need to make this larger as well right?

src/Views/EncryptView.vala
zeebok: Gotcha, I saw a couple other views use Dialog and override the size. Should I fix those as well and just rename this to icon clean-up?

src/Views/EncryptView.vala
```diff
@@ -32,9 +32,13 @@ public class EncryptView : AbstractInstallerView {
     }
 
     construct {
-        var image = new Gtk.Image.from_icon_name ("drive-harddisk", Gtk.IconSize.DIALOG);
+        var image = new Gtk.Image.from_icon_name ("drive-harddisk", Gtk.IconSize.INVALID) {
+            pixel_size = 128,
+            valign = Gtk.Align.END
```
danirabbit, line 37: afaict this valign isn't doing anything here?

src/Views/EncryptView.vala
danirabbit: If you want to change the others too sure, but I'm not super worried about it since this property goes away in Gtk4 anyways

src/Views/EncryptView.vala
zeebok: Hmmm not that I can visibly tell

src/Views/EncryptView.vala
zeebok: Certain ones need the align and some don't so I'll leave it

danirabbit (approved):
Looks good, thank you!

PR #724, opened 2023-07-29
daemon: Deactivate LVM devices before install

davidmhewitt:
I think this may fix the failure to clean install over an existing installation issue but I'll test this out in a VM now before marking it ready


davidmhewitt:
Okay, it doesn't fix it because this `distinst` method only deactivates encrypted logical devices, not all logical devices. I'll look at it some more.

PR #737, opened 2023-11-09
Prepare support for other backends

meisenzahl:
This serves as preparation to be able to support other installer backends.


meisenzahl:
I successfully tested this with elementaryos-8.0-daily.20240925.iso.

<details>
<summary>Steps to test this</summary>
- Try Demo Mode
- Add # to deb cdrom in /etc/apt/sources.list
- sudo apt update
- sudo apt install git
- git clone https://github.com/elementary/installer
- cd installer
- git checkout prepare-backends
- sudo apt build-dep io.elementary.installer
- meson build --prefix=/usr
- ninja -C build
- sudo ninja -C build install
- sudo systemctl daemon-reload
- sudo systemctl restart io.elementary.InstallerDaemon
- Log Out
- Install as usual
</detail>

common/DBusStructures.vala
```diff
@@ -95,7 +139,8 @@ public struct InstallerDaemon.InstallConfig {
     string keyboard_layout;
     string keyboard_variant;
     string lang;
-    uint8 flags;
+    bool modify_boot_order;
```
danirabbit, line 142: Is there a reason you chose not to use flags here?

src/Utils.vala
```diff
@@ -264,13 +264,6 @@ namespace Utils {
         string hostname = get_ubiquity_compatible_hostname () ?? ("elementary-os" + "-" + get_chassis ());
         hostname += "-" + get_machine_id ().substring (0, 8);
 
-        // If the automatic hostname logic fails in some way, it's possible we may generate an invalid
-        // hostname. We could fix this by trimming traling/leading hyphens or other invalid characters.
-        // But it's probably a bad hostname anyway, so just fallback
-        if (!Distinst.validate_hostname (hostname)) {
-            hostname = "elementary-os";
-        }
```
danirabbit, line 272: Did this move somewhere else? It seems like a good thing to have some kind of validation here

danirabbit (commented):
A couple questions :)

src/Utils.vala
meisenzahl: Since the validation is only available with distinst, I ported the logic with the last commit and took over the tests from distinst.

src/HostnameValidator.vala
```diff
@@ -0,0 +1,47 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * Authored by: Marius Meisenzahl <mariusmeisenzahl@gmail.com>
+ */
+
+namespace Utils {
+    private bool hostname_is_valid_char (char c) {
+        return ((c >= 'a' && c <= 'z') ||
+                (c >= 'A' && c <= 'Z') ||
+                (c >= '0' && c <= '9') ||
```
danirabbit, line 24: Is this just https://valadoc.org/glib-2.0/char.isalnum.html ?

src/HostnameValidator.vala
```diff
@@ -0,0 +1,47 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * Authored by: Marius Meisenzahl <mariusmeisenzahl@gmail.com>
+ */
```
danirabbit, line 18: Since this is based on MIT code, I don't think we can just change the license here to GPL. System76's copyright must also be maintained. Let's also use the compact SPDX header for new classes

src/HostnameValidator.vala
```diff
@@ -0,0 +1,47 @@
...
+    private bool hostname_is_valid_char (char c) {
+        return ((c >= 'a' && c <= 'z') ||
+                (c >= 'A' && c <= 'Z') ||
+                (c >= '0' && c <= '9') ||
+                c == '-' ||
+                c == '.');
+    }
+
+    // Based on https://github.com/pop-os/hostname-validator/blob/458fa5a1df98cb663f0456dffb542e1a907861c9/src/lib.rs#L29
+    public bool hostname_is_valid (string hostname) {
```
danirabbit, line 30: Should retain this original comment here which is pretty useful imo:

```
/// Validate a hostname according to [IETF RFC 1123](https://tools.ietf.org/html/rfc1123).
///
/// A hostname is valid if the following condition are true:
///
/// - It does not start or end with `-` or `.`.
/// - It does not contain any characters outside of the alphanumeric range, except for `-` and `.`.
/// - It is not empty.
/// - It is 253 or fewer characters.
/// - Its labels (characters separated by `.`) are not empty.
/// - Its labels are 63 or fewer characters.
/// - Its labels do not start or end with '-' or '.'.
```

danirabbit (changes requested):

danirabbit:
@meisenzahl Would it be too much to ask to propose the hostname validator bit as a separate branch? I think that's probably worth its own commit

danirabbit (approved):
This looks good to me! I can confirm that the new DistinstBackend file is basically what was in `Daemon` plus a few methods to convert from Distinst structs to our internal structs

PR #749, opened 2024-02-14
Port to GTK 4

### LINKED ISSUE #628: Port Installer to GTK4

danirabbit:
Fixes #628


danirabbit:
Tested the constraint layout with a number of different partitions with small and large partitions on either end and with unused space:

Screenshot from 2024-06-06 16 54 19

ryonakano:
We also need to update the dependencies in README :)

danirabbit:
@ryonakano done!

zeebok:
Just tried testing this and noticed this missing icon? 
Screenshot from 2024-07-27 20 28 06

zeebok (approved):
Nothing in the code sticks out, and running in test mode worked well except for some reason trying to modify partitions just opened gparted on my machine. But I assume that is just from me running it locally instead of in a normal live session or in VM

danirabbit:
@zeebok ah yeah "modify" always opens gparted. There's no partition resizing here. You can only choose what to apply on each partition but can't resize or create or delete partitions

PR #752, opened 2024-02-14
GTK4 Prep

danirabbit:
Easy GTK4 prep broken out into commits that should make it easier to review. Can be rebase merged if desired :shrug:


zeebok (approved):
This looks good. Testing worked as well.

PR #761, opened 2024-05-01
Remove builtin flag from installer session

davidmhewitt:
I believe this is the cause of the crashes in the live session in the daily ISOs.

I haven't looked to see if we also need to do some systemd stuff as well as this, but removing this from the ISO at least makes the crashing stop.


lenemter (approved):
:tada:

PR #765, opened 2024-05-07
PartitionBar: make sure we get click events

### LINKED ISSUE #763: Unable to select partition in partition bar
### What Happened?

Latest daily build installer partition bar in custom mode I'm unable to select any existing partition to install to, neither able to pick the efi partition. When I click on the partition nothing happens. Anyone else see this?

Perhaps https://github.com/elementary/installer/pull/751 is involved?

### Steps to Reproduce

Boot up current os8 daily ISO and go into custom partitioning to install to existing partitions and also select existing EFI ESP to mount where this issue occurs.

### Expected Behavior

Should be able to choose from the partition bar.

### OS Version

8.x (Early Access)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

danirabbit:
Fixes #763
Closes #767 

I can't actually get the menu to popup in this branch or any previous commit since the last release, but this makes sure the gesture sends the event at least. Maybe the menu only works in an actual installer session for some reason?

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -13,7 +13,7 @@ gnome = import('gnome')
 i18n = import('i18n')
 
 distinst_dep = dependency('distinst')
-glib_dep = dependency('glib-2.0')
+glib_dep = dependency('glib-2.0', version: '>=2.74')
 gobject_dep = dependency('gobject-2.0')
 gtk_dep = dependency('gtk+-3.0')
 gee_dep = dependency('gee-0.8')
--- a/src/Views/KeyboardLayoutView.vala
+++ b/src/Views/KeyboardLayoutView.vala
@@ -186,7 +186,7 @@ public class KeyboardLayoutView : AbstractInstallerView {
 
         show_all ();
 
-        Idle.add (() => {
+        Idle.add_once (() => {
             string? country = Configuration.get_default ().country;
             if (country != null) {
                 string default_layout = country.down ();
--- a/src/Widgets/PartitionBar.vala
+++ b/src/Widgets/PartitionBar.vala
@@ -5,7 +5,7 @@
  * Authored by: Michael Aaron Murphy <michael@system76.com>
  */
 
-public class Installer.PartitionBar : Gtk.Box {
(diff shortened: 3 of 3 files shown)


vjr:
@danirabbit 

When I click I see a message like "gtk popover - assertion failed - GTK_IS_POPOVER is not a popover"...

I had tried to fix this issue - I also changed Box to EventBox and also re-added the button pressed mask. I had to move the `gesture.connect (menu.popup)` from the construct block up to the constructor, because apparently the block was being executed before the constructor where the `menu` was actually being created - this fixed the debug message about "GTK_IS_POPOVER not being a popover".

I also re-added the local/private method that called `menu.popup ()` with a log statement that confirmed the click event is happening but `menu.popup ()` does nothing.

Is there some sort of GTK setup required - like adding the widgets in a hierarchy so that the event from the eventbox (partitionbar) is propagated down to the partitionmenu widget?

hope what ive written here is not too much gibberish :-)

danirabbit:
@vjr okay so this has multiple layers haha.

One layer is this: which means even when the menu does popup it's closed immediately https://github.com/elementary/installer/pull/767

vjr:
@danirabbit so how do i test? do i need to apply both PRs or just #767 ?

danirabbit:
@vjr At the very least we need #767, but I don't think the issue is fixed even with both of these applied. So to test #767 you can see that keyboard navigation is fixed after the keyboard layout selection view.

Then we have this issue which is the click event wasn't actually being sent, but with the eventbox it does send

And then I think there might be a third thing which is I don't know that the menu is actually having all its widgets shown properly, but I can't really investigate it further because of the bug from #767

danirabbit:
@vjr okay actually, it was simpler than I thought and it's now fixed in this branch. This can be rebase merged :)

vjr (approved):
LGTM - the popup is working now - please merge so the upcoming daily iso builds get this fix - i intend to install and further test os8 - thanks!

PR #766, opened 2024-05-07
AbstractInstallerView: replace ButtonBox with Boxes

danirabbit:
No ButtonBox in Gtk4, so use two boxes


zeebok:
This works but I do notice that some buttons are different sizes:
**Before**
Screenshot from 2024-05-23 14 16 37

**After**
Screenshot from 2024-05-23 14 16 21

PR #776, opened 2024-05-28
Application: modernize and set up locale

### LINKED ISSUE #775: Installer is not localized
### What Happened?

Installer of OS 8 Early Access does not get localized like 7.1 Horus.

### Steps to Reproduce

1. Boot from Early Access (I used `elementaryos-8.0-daily.20240525.iso`)
2. Wait until the installer is shown
3. See the languages list is not localized

image 5

4. Select some language other than English
5. See the Installer UI is not localized after navigating to keyboard layout selection page

image4

### Expected Behavior

1. In the step 3, the languages list should be localized like Horus:

image6

2. In the step 5, the Installer UI should be localized like Horus:

image3

### OS Version

8.x (Early Access)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

danirabbit:
Maybe fixes #775 ?

Modernize application class:
* Use built in way to load icons
* Set up everything locale related we do in other apps
* Set app ID and flags in object call
* load disk bar CSS in startup not activate

### DIFF
--- a/data/io.elementary.installer.gresource.xml
+++ b/data/io.elementary.installer.gresource.xml
@@ -8,4 +8,8 @@
     <file alias="io.elementary.installer.caution.svg" compressed="true" preprocess="xml-stripblanks">caution.svg</file>
     <file alias="io.elementary.installer.emblem-downloads.svg" compressed="true" preprocess="xml-stripblanks">emblem-downloads.svg</file>
   </gresource>
+  <gresource prefix="/io/elementary/installer/icons">
+    <file alias="io.elementary.installer.caution.svg" compressed="true" preprocess="xml-stripblanks">caution.svg</file>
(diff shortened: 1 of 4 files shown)


data/io.elementary.installer.gresource.xml
```diff
@@ -8,4 +8,8 @@
     <file alias="io.elementary.installer.caution.svg" compressed="true" preprocess="xml-stripblanks">caution.svg</file>
     <file alias="io.elementary.installer.emblem-downloads.svg" compressed="true" preprocess="xml-stripblanks">emblem-downloads.svg</file>
```
ryonakano, line 9: You moved these icons under the /io/elementary/installer/icons prefix, so maybe we can remove these lines?

danirabbit: Oops yes you're right. Thanks for catching that!

ryonakano (approved):

PR #780, opened 2024-05-28
Terminal: gtk4prep

danirabbit:
* Can't subclass scrolledwindow
* Code style
* Can't connect to size_allocate anymore, so just try to scroll on idle
* Add log output to test mode so we can test it


zeebok:
I am not sure if it is because of the code changes or something else but when I get to the progress view, I see the "I'm faking it!" log message but then the app crashes with 

`Bail out! Gtk:ERROR:../../../gtk/gtktextview.c:4452:gtk_text_view_validate_onscreen: assertion failed: (priv->onscreen_validated)`

danirabbit:
@zeebok How about now?

zeebok (approved):

PR #782, opened 2024-05-29
AbstractInstallerView: DRY more layout

danirabbit:
Same as we do in Initial Setup, have the Abstract View handle more of the layout for each pane. Removes a lot of redunant code, will make it easier to style with CSS in GTK4, and make sure we're being consistent between views

The biggest change is for the partition view, everything else pretty much looks the same:
Screenshot from 2024-05-29 10 09 16


src/Views/ProgressView.vala
```diff
@@ -15,7 +15,7 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-public class ProgressView : AbstractInstallerView {
+public class ProgressView : Gtk.Box {
```
zeebok, line 18: Why is this being changed away from AbstractInstallerView?

danirabbit: Because it's the only pane that isn't a side-by-side layout and the abstract view now enforces the side-by-side layout

zeebok (approved):

Issue #793, opened 2024-08-25
Clicking a partition in the custom install glitches the installer window


ryonakano:
### What Happened?

Clicking a partition in the custom install glitches the installer window, which makes me being unable to set up partitions.

### Steps to Reproduce

1. Boot from Early Access 8 ISO (I used `elementaryos-8.0-daily.20240824.iso`)
2. Select the custom install
3. In the "Select Partitions" view, click "Modify Partitions…" to create partitions if none exists
4. In the "Select Partitions" view, click the storage bar (the yellow/green portion in the following screenshot)  
![VirtualBox_test_25_08_2024_10_52_43](https://github.com/user-attachments/assets/154a3849-fe8b-48a3-bc93-fb396f8c9af6)
5. See the installer window get bigger and unable to setup partitions  
![VirtualBox_test_25_08_2024_10_54_19](https://github.com/user-attachments/assets/d6cbd548-3471-459d-8633-98a8567eee02)

### Expected Behavior

The window should not glitch and the following popover should be shown where we can setup partitions:

![VirtualBox_test_25_08_2024_10_57_47](https://github.com/user-attachments/assets/9dcc125a-b4a7-4bfb-bfcc-ae25816f0dfc)

### OS Version

8.x (Early Access)

### Software Version

Latest release (I have run all updates)

### Log Output

```shell
Aug 25 02:09:05 elementary io.elementary.g[1858]: Can't update stage views actor unnamed [MetaWindowActorX11] is on because it needs an allocation.
Aug 25 02:09:05 elementary io.elementary.g[1858]: Can't update stage views actor unnamed [MetaSurfaceActorX11] is on because it needs an allocation.
Aug 25 02:09:05 elementary io.elementary.i[2053]: gdk_monitor_get_geometry: assertion 'GDK_IS_MONITOR (monitor)' failed
Aug 25 02:09:05 elementary io.elementary.i[2053]: gdk_monitor_get_geometry: assertion 'GDK_IS_MONITOR (monitor)' failed
Aug 25 02:09:05 elementary io.elementary.i[2053]: gdk_monitor_get_geometry: assertion 'GDK_IS_MONITOR (monitor)' failed
Aug 25 02:09:05 elementary io.elementary.i[2053]: gtk_widget_size_allocate(): attempt to allocate GtkPopoverContent contents 0x562c36a38a20 with width 253 and height -23
[...]

PR #802, opened 2024-09-12
MainWindow: use Adw.NavigationView

danirabbit:
Replace deprecated Leaflet with NavigationView.

Since NavigationView doesn't have a property for disabling navigation, we replace the navigation stack to prevent navigation. There's some tricks with waiting for `shown` so that we still get animations


danirabbit:
@ryonakano Can you see if this fixes #793 for you? I'm thinking it might be a bug with Leaflet on X11

ryonakano:
> @ryonakano Can you see if this fixes #793 for you? I'm thinking it might be a bug with Leaflet on X11

Tried Testing Mode on X11, but unfortunately not:

expand

[test-screen0.webm](https://github.com/user-attachments/assets/867b4ce3-4ffd-4791-8a68-26ad2dc19ebe)

src/MainWindow.vala
```diff
@@ -128,136 +125,138 @@ public class Installer.MainWindow : Gtk.Window {
...
+        disk_view.cancel.connect (() => navigation_view.pop ());
+        disk_view.next_step.connect (() => load_encrypt_view ());
     }
 
-    private void load_check_view () {
-        if (check_ignored) {
-            return;
-        }
-
+    private void load_check_view (string path) {
```
ryonakano, line 166: Is there any reason you chose `string` to separate standard install and custom install? I think we may want to use a enum or bool which are more easier to handle.

src/Views/DriversView.vala
```diff
@@ -75,7 +73,7 @@
         content_area.append (message_box);
 
         var back_button = new Gtk.Button.with_label (_("Back"));
-        back_button.clicked.connect (() => ((Adw.Leaflet) get_parent ()).navigate (BACK));
+        back_button.clicked.connect (() => ((Adw.NavigationView) get_parent ()).pop ());
```
ryonakano, line 76: Honestly I think the each AbstractInstallerView shouldn't know that its parent is Adw.NavigationView; it should just emit a signal like `previous_step ()`, connect to the signal in MainWindow, and call `Adw.NavigationView.pop ()` in its handler. But that's out of scope of this PR, so leaving just a comment.

ryonakano (commented):
Confirmed all pages transit to the expected page and back to the previous page.

Code almost looks good, but leaving some comments.

src/Views/DriversView.vala
danirabbit: Made it into an action :)

src/MainWindow.vala
danirabbit: Re-wrote it without the strings!

ryonakano (approved):
Let's go :rocket:

PR #813, opened 2024-09-24
PartitionBlock: GObject

danirabbit:
Moves the menu up to partitionview so we're not passing it through two objects


ryonakano (commented):
Maybe the following signal no longer used?

```
src/Widgets/PartitionBlock.vala:9:    public signal void decrypted (InstallerDaemon.LuksCredentials credential);
```

ryonakano (approved):

PR #815, opened 2024-09-24
PartitionMenu: write check_values as single function with early returns

danirabbit:
Also make sure we're updating icon and add warning text for invalid custom value while we're here

### DIFF
--- a/src/Widgets/PartitionMenu.vala
+++ b/src/Widgets/PartitionMenu.vala
@@ -278,6 +278,7 @@ public class Installer.PartitionMenu : Gtk.Popover {
             ));
 
             partition_bar.icon = new ThemedIcon ("process-completed-symbolic");
+            partition_bar.tooltip_text = null;
         } catch (GLib.Error e) {
             partition_bar.icon = new ThemedIcon ("dialog-warning-symbolic");
             partition_bar.tooltip_text = e.message;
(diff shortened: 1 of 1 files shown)


src/Widgets/PartitionMenu.vala
```diff
@@ -250,39 +250,36 @@ public class Installer.PartitionMenu : Gtk.Popover {
...
-        partition_bar.icon = new ThemedIcon (
-            error == null ? "process-completed-symbolic" : "dialog-warning-symbolic"
-        );
 
-        if (error != null) {
-            partition_bar.tooltip_text = error;
+            partition_bar.icon = new ThemedIcon ("process-completed-symbolic");
+        } catch (GLib.Error e) {
+            partition_bar.icon = new ThemedIcon ("dialog-warning-symbolic");
+            partition_bar.tooltip_text = e.message;
```
ryonakano, line 282: It looks like the tooltip text remains after resolving the error, e.g. change not to use the partition or change to valid usage. This also happens on main so maybe we may want to fix it in another PR?

VirtualBox_elementary OS 8_01_10_2024_23_07_16

VirtualBox_elementary OS 8_01_10_2024_23_07_37

danirabbit: Fixed!

ryonakano (approved):

PR #819, opened 2024-09-30
Launcher: use dbus to log out

### LINKED ISSUE #806: Installer doesn't launch from Applications menu in Circe daily
### What Happened?

On a live USB boot of OS 8 daily, selecting the installer option from Applications menu does nothing.

But launching `io.elementary.installer` from CLI does launch it, albeit from the very start (until later running into #793).

### Steps to Reproduce

1. Boot up the daily from a USB.
2. Try demo mode. 
3. Open Applications menu
4. Click on installer.

Nothing happens.

### Expected Behavior

It launches and starts asking stuff.

### OS Version

8.x (Early Access)

### Software Version

### Log Output

_No response_

### Hardware Info

Ye olde Dell Vostro 2520 with i5 gen 3

danirabbit:
Fixes #806 

confirmed that this command works on latest RC

### DIFF
--- a/data/io.elementary.installer.desktop.in
+++ b/data/io.elementary.installer.desktop.in
@@ -1,7 +1,7 @@
 [Desktop Entry]
 Name=Install elementary OS
 Comment=Install the system
-Exec=gnome-session-quit --logout
+Exec=dbus-send --session --dest=org.gnome.SessionManager --print-reply /org/gnome/SessionManager org.gnome.SessionManager.Logout uint32:0
 Icon=io.elementary.installer
 Terminal=false
 Type=Application


meisenzahl (approved):

ryonakano:
`gnome-session-quit --logout --no-promt` seems to work as expected but this PR is merged so it's fine anyways

ryonakano:
@danirabbit

> confirmed that this command works on latest RC

How did you confirmed this? Calling the dbus-send in the live session does nothing for me……

VirtualBox_test_01_10_2024_21_23_54

meisenzahl:
I have to admit that I have not tested this in an installer session. I am sorry.

danirabbit:
@ryonakano I tested it in demo mode yesterday just like that. Dunno why that didn't work for you 🤔

ryonakano:
Hmm, after I trigger the logout dialog from Wingpanel, both `gnome-session-quit` and `dbus-send` began to work as expected:

https://github.com/user-attachments/assets/8a0a402e-7b9c-4273-abc8-e940ae14f1ae

danirabbit:
I just booted the latest daily and it does show the log out dialog here when clicking the installer launcher

meisenzahl:
I have tested the behavior with elementaryos-8.0-daily.20241003.iso and elementaryos-8.0-daily.20241004.iso. Unfortunately, I cannot confirm that the behavior works as desired.

My behavior is the same as @ryonakano's.

However, when I enter `gnome-session logout` in Terminal, I am immediately logged out without a dialog. Can you confirm this behavior? Could this be the right command?

ryonakano:
> However, when I enter `gnome-session logout` in Terminal, I am immediately logged out without a dialog. Can you confirm this behavior? Could this be the right command?

I can confirm that behavior, but on OS 7 the logout dialog is prompt before actual logout

leolost2605:
Just FYI `gnome-session-quit --no-prompt` will log out instantly without going the route via the session manager and therefore wingpanel or anything. If all else fails this might be the most robust way. 
At least that's how it works in the normal session not sure whether anything changes here but I guess it shouldn't?

PR #821, opened 2024-10-02
Send battery notification instead of infobar

danirabbit:
Since we have notifications in the installer session now. Doesn't block UI and fixes a deprecation warning


ryonakano:
Tested on my old MacBook Air, and confirmed the notification appears when unplugged and disappears when plugged.

One thing I'm concerning is that the notification body may be ellipsized in some languages, although users still should be able to understand what they should do from the notification title:

image

danirabbit:
@ryonakano made the body text a lot shorter. I suppose we probably need to open an issue in Notifications about how to handle longer strings

ryonakano:
@danirabbit Thanks for the fix. Handling notifications with long text seems to be already reported at https://github.com/elementary/notifications/issues/66 so I approve this now :)

ryonakano (approved):

PR #823, opened 2024-10-02
Move session files to their own folder

danirabbit:
Separate installer application data from session data

PR #824, opened 2024-10-02
Session: build GSD autostarts

danirabbit:
Builds on #823 

Create GSD autostarts in the build system like we do with our regular sessions


ryonakano:
@danirabbit

> Dunno why CI is unhappy 🤔 builds locally

Opened #827 to make CI happy :)

README.md
```diff
@@ -14,6 +14,8 @@ You'll need the following dependencies:
  - meson
  - desktop-file-utils
  - gettext
+ - gnome-settings-daemon
+ - gnome-settings-daemon-dev
```
ryonakano, line 18:
```suggestion
 - gnome-settings-daemon-dev >= 3.36.0
```

ryonakano (changes requested):
Tested this branch by building an ISO. The volume up & volume down media keys does not work in the Installer Greeter, while it works with latest RC image (`elementaryos-8.0-stable.20241001rc.iso`).

PR #826, opened 2024-10-02
HostnameValidatorTest: cleanup, comments, more tests

danirabbit:
HostnameValidator.vala:
- Add comments inline and separate lines so we know exactly what each line tests for
- Use `has_prefix` and `has_suffix`

HostnameValidatorTest:
- Give this test a unique name for CI
- Make the test one file
- Fix path syntax in meson
- Add comments so we know what each line tests for
- Make sure we test for all of the things that the function should be checking, notably empty hostname and a super long hostname with otherwise short labelsni

### DIFF
--- a/src/HostnameValidator.vala
+++ b/src/HostnameValidator.vala
@@ -4,40 +4,50 @@
  * SPDX-License-Identifier: MIT
  */
 
-namespace Utils {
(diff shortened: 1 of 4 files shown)


meisenzahl (approved):
Looks good! Thanks for the cleanup 😄

PR #827, opened 2024-10-05
meson: Set prefix to /usr

ryonakano:
The following error message in the CI means the sysconfdir variable is set to `etc`:

    session/meson.build:61:4: ERROR: File etc/xdg/autostart/org.gnome.SettingsDaemon.A11ySettings.desktop does not exist.

It points to `/etc` only when the prefix is set to `/usr` according to https://mesonbuild.com/Builtin-options.html:

    When the prefix is /usr: sysconfdir defaults to /etc, localstatedir defaults to /var, and sharedstatedir defaults to /var/lib

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,7 +28,7 @@ jobs:
       env:
         DESTDIR: out
       run: |
-        meson setup build
+        meson setup build --prefix=/usr
         ninja -C build install
 
   lint:


session/meson.build
```diff
@@ -52,11 +52,12 @@ install_data(
     install_dir: schemadir
 )
 
+vendor_prefix = 'pantheon'
```
danirabbit, line 55: We need to use a different prefix here or this package will conflict with our other session package

Issue #828, opened 2024-10-06
Canadian french keyboard is offered after selecting Canadian English


eucharistia:
### What Happened?

When I install the Sponsor RC for eOS 8, I select Canadian English and then I'm offered the french Canadian keyboard (likely because it's the only Canadian keyboard)

### Steps to Reproduce

1. Attempt to install eOS 8
2. Select English
3. Select Canada
4. Be offered the french Canadian keyboard

### Expected Behavior

1. Select English
2. Select Canada
3. Be offered the standard US keyboard

### OS Version

8.x (Early Access)

### Session Type

Classic Session (X11, This is the default)

### Software Version

Latest release (I have run all updates)

ryonakano:
> (likely because it's the only Canadian keyboard)

This seems to be correct. We set the layout as default whose name matches with the user chosen country here:

https://github.com/elementary/installer/blob/4ad60d99515762b4623fe3b1b41634af274258a7/src/Views/KeyboardLayoutView.vala#L169-L188

And looking at `/usr/share/X11/xkb/rules/base.xml` which we parse to get the list of layouts the country name of Canadian `ca` matches `French (Canada)` keyboard:

```xml
    <layout>
      <configItem>
        <name>ca</name>    ←Current Layout
        <!-- Keyboard indicator for French layouts -->
        <shortDescription>fr</shortDescription>
        <description>French (Canada)</description>
        <countryList>
          <iso3166Id>CA</iso3166Id>
        </countryList>
        <languageList>
          <iso639Id>fra</iso639Id>
        </languageList>
      </configItem>
      <variantList>
        <variant>
          <configItem>
            <name>fr-dvorak</name>
            <!-- Keyboard indicator for French layouts -->
            <shortDescription>fr</shortDescription>
            <description>French (Canada, Dvorak)</description>
          </configItem>
        </variant>
        <variant>
          <configItem>
            <name>fr-legacy</name>
            <!-- Keyboard indicator for French layouts -->
            <shortDescription>fr</shortDescription>
            <description>French (Canada, legacy)</description>
          </configItem>
        </variant>
        <variant>
          <configItem>
            <name>multix</name>
            <description>Canadian (CSA)</description>
          </configItem>
        </variant>
        <variant>
          <configItem>
            <name>eng</name>    ←Expected Layout
            <!-- Keyboard indicator for English layouts -->
            <shortDescription>en</shortDescription>
            <description>English (Canada)</description>
            <languageList>
[...]

PR #833, opened 2024-10-09
KeyboardLayout: list from xkbregistry

### LINKED ISSUE #652: Use libxkbregistry
We are currently parsing the keyboard information ourselves. Since xkbcommon 1.0, the xkbregistry library does it for us, we should use it instead.

Documentation: https://xkbcommon.org/doc/current/group__registry.html

danirabbit:
Fixes #652


ryonakano:
Wow, does this fix #828 hopefully?

danirabbit:
@ryonakano I don't think it does. That was my original interest in it though!

src/Objects/KeyboardLayout.vala
```diff
@@ -246,79 +246,31 @@ public class Installer.KeyboardLayout : GLib.Object {
...
-            unowned Xml.Node* config_node = get_xml_node_by_name (layout_node, "configItem");
-            unowned Xml.Node* variant_node = get_xml_node_by_name (layout_node, "variantList");
-            unowned Xml.Node* description_node = get_xml_node_by_name (config_node, "description");
-            unowned Xml.Node* name_node = get_xml_node_by_name (config_node, "name");
-            if (name_node == null || description_node == null) {
+        var xkb_layout = xkb_context.get_first_layout ();
+        while (xkb_layout != null) {
+            if (xkb_layout.get_variant () != null) {
                 continue;
             }
```
ryonakano, line 259: This looks like an infinite loop if `xkb_layout.get_variant () != null` because the value of `xkb_layout` wouldn't change here

danirabbit: Ah yeah that was just kind of leftovers from trying to figure out how to handle variants. Thanks for catching that!

ryonakano (commented):
One small comment.

src/Objects/KeyboardLayout.vala
```diff
@@ -246,79 +246,27 @@ public class Installer.KeyboardLayout : GLib.Object {
         return a.display_name.collate (b.display_name);
     }
 
-    private const string XKB_RULES_FILE = "base.xml";
-
-    private static string get_xml_rules_file_path () {
-        unowned string? base_path = GLib.Environment.get_variable ("XKB_CONFIG_ROOT");
-        if (base_path == null) {
-            base_path = Build.XKB_BASE;
```
ryonakano, line 254: Ah, maybe we can ditch this constant:

```
[ryo@b760m ~/work/elementary/installer (danirabbit/rxkb-layouts =)]$ git grepn XKB_BASE
src/Config.vala.in:6:    public const string XKB_BASE = "@XKB_BASE@";
src/meson.build:44:configuration_data.set('XKB_BASE', xkb_base)
[ryo@b760m ~/work/elementary/installer (danirabbit/rxkb-layouts =)]$
```

danirabbit: Good catch!

ryonakano (approved):
Nice improvements!

PR #834, opened 2024-10-11
Gsettings Overrides: add installer to dock

danirabbit:
Fixes https://github.com/elementary/dock/issues/239


danirabbit:
Nevermind, this is for installer session not demo mode which is a regular pantheon session :sweat_smile:

PR #836, opened 2024-10-14
Application: Initialize translation at first

tintou:


ryonakano:
Am I right that this is meant to fix #775?

tintou:
I'm not sure this will fix this but it is more correct anyway to do it before creating the GApplication

danirabbit (approved):

PR #839, opened 2024-10-17
Launcher: use gnome-session-quit --no-prompt

### LINKED ISSUE #806: Installer doesn't launch from Applications menu in Circe daily

danirabbit:
Fixes #806
Closes #820

Sounds like this solution still isn't working for some folks, but this one should for sure work? It works in a VM here at least

### DIFF
--- a/data/io.elementary.installer.desktop.in
+++ b/data/io.elementary.installer.desktop.in
@@ -1,7 +1,7 @@
 [Desktop Entry]
 Name=Install elementary OS
 Comment=Install the system
-Exec=dbus-send --session --dest=org.gnome.SessionManager --print-reply /org/gnome/SessionManager org.gnome.SessionManager.Logout uint32:0
+Exec=gnome-session-quit --no-prompt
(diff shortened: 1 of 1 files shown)


ryonakano (approved):
Tested by building an ISO with this branch and confirmed now it logouts when I click the Installer icon in Applications Menu. I'm :+1: if it's fine now it logouts immediately without the logout confirmation dialog.

PR #843, opened 2024-10-19
Set LANG variable on startup

### LINKED ISSUE #775: Installer is not localized

davidmhewitt:
Fixes #775

gettext doesn't give us translations if `LANG` is set to `C.UTF-8` (which it is initially on the ISO)

If we initially set this to English, gettext initializes properly and then we can correctly change the language at runtime with the `LANGUAGE` variable after that.

### DIFF
--- a/src/Application.vala
+++ b/src/Application.vala
@@ -79,6 +79,14 @@ public class Installer.App : Gtk.Application {
 }
 
 public static int main (string[] args) {
+    // When the installer starts on a live iso session, LANG is set to C.UTF-8, which causes gettext to ignore the
+    // LANGUAGE variable. To enable runtime language switching, we need to set LANG to something other than C.UTF-8. See:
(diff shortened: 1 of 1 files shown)


ryonakano (approved):
Build the ISO and confirmed the issue now fixed finally! Thank you very much for your investigation :heart:

VirtualBox_test_19_10_2024_15_41_40

VirtualBox_test_19_10_2024_15_43_36

PR #874, opened 2025-04-14
debian: allow all architectures

NN708:
This PR allows building the installer on all architectures.

Related to elementary/os#754. Tested on ARM64 server.


danirabbit:
Let's give it a go 🚀

PR #884, opened 2025-06-30
MainWindow: center on X11

### LINKED ISSUE #881: Installer UI is not centered
### What Happened?

While installing a daily ElementaryOS build (elementaryos-8.0-daily.20240328.iso) the installer's UI is not centered. I had the same issue months ago with a stable version of ElementaryOS 7, however I did not record it at that time. 

### Steps to Reproduce

1) Flash ElementaryOS iso to a USB drive with balena etcher.
2) Boot into the boot menu and select the default install/try ElementaryOS (not safe mode).
IMG_20240407_152121_144

### Expected Behavior

I am fairly certain the UI is supposed to be centered and not off to one side.

### OS Version

8.x (Early Access)

### Software Version

### Log Output

_No response_

### Hardware Info

I am using an Asus ROG Zephyrus G15 laptop that has an RTX 3060 Laptop Gpu alongside the integrated graphics in its Ryzen 6900HS.

danirabbit:
Fixes #881


stsdc:
Just a thought: I see a lot of these if (wayland) { } else { do_xserver}.
Maybe we can move that out to the `pantheon_wayland`?
When the time for dropping the X11 will come, it will be less work, no?

danirabbit:
@stsdc Maybe but I think that's outside the scope of this pull request :)

@jeremypw an easy way to test is to resize the window and see if it remains centered when resizing

src/MainWindow.vala
```diff
@@ -95,10 +95,29 @@ public class Installer.MainWindow : Gtk.ApplicationWindow, PantheonWayland.Exten
...
+    private void make_centered_x11 () {
+        var display = Gdk.Display.get_default ();
+        if (display is Gdk.X11.Display) {
+            unowned var xdisplay = ((Gdk.X11.Display) display).get_xdisplay ();
+
+            var window = ((Gdk.X11.Surface) get_surface ()).get_xid ();
+
+            var prop = xdisplay.intern_atom ("_MUTTER_HINTS", false);
+
+            var value = "centered=1:restore-previous-region=0";
```
lenemter, line 115: You don't need to use restore-previous-region here

```suggestion
            var value = "centered=1";
```

PR #887, opened 2025-07-06
Use CSS for tnum instead of Pango

danirabbit:
Avoid the screen reader reading markup by using CSS


src/Views/ProgressView.vala
```diff
@@ -240,7 +241,7 @@ public class ProgressView : Adw.NavigationPage {
                     break;
             }
 
-            progressbar_label.label = "<span font-features='tnum'>%s (%d%%)</span>".printf (step_string, status.percent);
```
tintou, line 243: Have you tried using progressbar_label.set_markup ?

danirabbit: As far as I can tell, the screen reader will still read out pango markup tags even when using `set_markup` instead of `label`

tintou: have you tried it? (because set_markup is following a different code path than just setting the label)

lenemter (approved):
LGTM

PR #897, opened 2025-07-26
EncryptView: make warnings messages screen reader accessible

danirabbit:
Makes messages able to be navigated and read by the screen reader. Also add a key controller so that you can still start typing right away to fill in your encryption password without having to manually focus the entry first

### DIFF
--- a/src/Views/EncryptView.vala
+++ b/src/Views/EncryptView.vala
@@ -159,7 +159,7 @@ public class EncryptView : AbstractInstallerView {
         // Forward keys so we can type immediate when list is focused
         var key_controller = new Gtk.EventControllerKey ();
         key_controller.key_pressed.connect ((keyval, keycode, state) => {
-            if (keyval <= 65000) {
+            if (Gdk.keyval_to_unicode (keyval) != 0) {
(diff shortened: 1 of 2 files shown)


ryonakano (approved):

src/Views/EncryptView.vala
```diff
@@ -148,6 +155,21 @@ public class EncryptView : AbstractInstallerView {
             confirm_entry.is_valid = confirm_password ();
             update_next_button ();
         });
+
+        // Forward keys so we can type immediate when list is focused
+        var key_controller = new Gtk.EventControllerKey ();
+        key_controller.key_pressed.connect ((keyval, keycode, state) => {
+            if (keyval <= 65000) {
```
lenemter, line 162: What is this number? Shouldn't it be 65535? Maybe move it to constant

danirabbit: Do you know of a good existing constant for this? I was trying to find the line between navigation keys and characters

danirabbit: That's way better, thank you!

Issue #898, opened 2025-07-30
Installer crashes when navigating to the custom installation view


ryonakano:
### What Happened?

Installer crashes when navigating to the custom installation view on a system with partition already setup.

This sounds like a reappearance of #848.

### Steps to Reproduce

1. Install OS 8 using the entire of the disk ("Erase Disk and Install" option) and reboot
2. After installation, reboot to the installation ISO and go to the custom install view
3. See the installer crashes soon after navigating to the custom installation view

### Expected Behavior

Installer shouldn't crash.

### OS Version

8.x (Early Access)

### OS Architecture

arm64 (on Raspberry Pi, Apple Silicon, etc.)

### Session Type

Classic Session (X11, This is the default)

### Software Version

Latest release (I have run all updates)

### Log Output

```shell

```

danirabbit:
@ryonakano can you list the disks and partitions you are able to reproduce the issue with?

iyatskiv:
I do face the very same issue. I ran `busctl call io.elementary.InstallerDaemon /io/elementary/InstallerDaemon io.elementary.InstallerDaemon GetDisks "b" true` with next result
```
(a(ssttbba(sitt(yt)s))a(ssttbba(sitt(yt)s))) 2 "JetFlash Transcend 64GB" "/dev/sda" 119808000 512 true true 2 "/dev/sda1" 0 2048 52633599 0 0 "" "/dev/sda2" 6 52633600 52699135 2 0 "" "Samsung SSD 990 EVO Plus 1TB" "/dev/nvme0n1" 1953525168 512 false false 0 0
```

I'm also getting this errors in console when `io.elementary.installer` crashes.
```
elementary@elementary:~$ sudo io.elementary.installer

** (io.elementary.installer:24325): WARNING **: 08:07:35.824: ExtendedBehavior.vala:71: Couldn't make centered: ExtendedBehavior surface was null. Did you forget to call connect_to_shell?

(io.elementary.installer:24325): Gtk-CRITICAL **: 08:07:43.782: INTERNAL: ratio == DBL_MAX in dual_optimize

(io.elementary.installer:24325): Gtk-CRITICAL **: 08:07:43.782: INTERNAL: invalid entry variable during pivot

(io.elementary.installer:24325): Gtk-CRITICAL **: 08:07:43.782: gtk_constraint_expression_change_subject: assertion 'new_subject != NULL' failed
Segmentation fault
```

Issue #879, opened 2025-05-19
installer crashes / blanks when detecting partitioning scheme


johnfk75:
### What Happened?

Hello,
my problem might be same as in #848, but I did not find a takeaway solution there, just an explanation.
I have a maybe messy partitioning configuration. There is a windows 11 installation I want to keep. Then there is a partition I want to use for elementary os, labelled /. And then there are a number of small partitions whose use I am not sure of; whether they belong to the windows installation or a former elementary os installation (which at some point refused to boot). But I did not dare delete any of them.
Anyway, the installer crashes when trying to detect these. This is reproducable.  I think it started crashing when I labelled one partition /. 
It proved possible to install another linux (siduction). But I'd rather go with elementary, so any help would be appreciated!
Kind regards, Felix

### Steps to Reproduce

1. booted elementary 8.0 iso usb stick, go through first stages of installer until detection of partitioning scheme
2. renamed one partition manually to /
3. since then the automatic detection crashes.

![Image](https://github.com/user-attachments/assets/7ad82a7f-22e6-4cc5-a87a-903df0e99fc1)

### Expected Behavior

installer continues it's work.

### OS Version

8.x (Circe)

### Session Type

Classic Session (X11, This is the default)

### Software Version

Latest release (I have run all updates)

### Log Output

```shell

```

vjr:
What I do first of all, before starting with any linux distro, is shrink my Windows partition (or if you have multiple partitions, like C:, D: etc) from within Windows itself using the Disk Management tool.

I then boot into elementary OS (make sure you download the latest ISO build from the main website elementary.io) installer and go into custom install, run the partitioning tool there, create another ESP (EFI system partition) with generous size (1 GB here) min is I think 256 or 512mb I dont remember.

Then I create my other partition(s) intended to be used as root ext4 - for more elaborate configs - you should be able to figure it out.

Then back to the installer, click the EFI/ESP and select "use this" and "format" (if you are using it for the first time, else DO NOT format the EFI/ESP) and mount to `/boot/efi`, then the other root ext4 click it and select "USE THIS" and "FORMAT" as ext4 - I've not tried others like btrfs or encryption etc.

The installation should go through now and when reboot you should get the GRUB boot selection menu with elementary and windows and any others in the list.

Good luck.

sudanchapagain:
[...]

PR #901, opened 2025-08-16
DiskBar: Add some safety checks to constraints calculations

lenemter:
Addresses https://github.com/elementary/installer/issues/879 and https://github.com/elementary/installer/issues/898

Shouldn't produce any side effects. Tested by putting random numbers into test partitions in `PartitioningView.get_test_diskinfo` and this appears to fix them. This is not really a good testing but worth seeing whether   this fixes issue with real partitions.


danirabbit (approved):
Tested with the new crash case Ryo added and I can confirm this fixes the crash

PR #902, opened 2025-08-17
PartitioningView: add new crash case

danirabbit:
Adds Ryo's new crash case from #898


lenemter (changes requested):
This crash case appears to have 18 EB of unused space..? 

image

zeebok:
This has conflicts but I also see the issue it fixes is already resolved. Is this PR outdated now?

PR #942, opened 2026-08-11
Followup for "Session: create installer systemd service"

ryonakano:
Additional work for #929 to work. This should also fixes the current Daily OS 9 iso directly goes to the demo session instead of the installer session.

## Changes summary
-  Use gala instead of the greeter compositor (which Danielle wrote as TODO in the above PR)
- Rename `io.elementary.installer.compositor.service` to `io.elementary.installer.compositor@wayland.service` to make sure ConditionEnvironment with unescaped instance name `%I` works
-  Merge with [gala@wayland.service.in](https://github.com/elementary/gala/blob/cb99c833b6bb3438331213abf6af21d046e624f6/data/gala%40wayland.service.in) because I guess it was based on [gala@x11.service.in](https://github.com/elementary/gala/blob/cb99c833b6bb3438331213abf6af21d046e624f6/data/gala%40x11.service.in)
- Fix missing session.conf
  - This was actually needed to make things work, as also written in the doc: https://gnome.pages.gitlab.gnome.org/gnome-session/re01.html
  - The content is based on `RequiredComponents` in `session/installer.session`
- Change relationship between `io.elementary.installer.target` and `io.elementary.installer.service` so that Installer launches when switching to the installer session
  - Before:  The .service is _WantedBy_ the .target 
  - After: The .target _Wants_ the .service

## Checklists
- [x] Clicking the Installer icon in the dock goes to the installer session and Installer appears

https://github.com/user-attachments/assets/3620f0f5-b7bf-46e9-b171-abeb1d6b4929

Note: You'll need https://github.com/elementary/gala/pull/2889 to make Wingpanel appear in the installer session.


danirabbit (approved):
Thanks for fixing this!

